PR #2, opened 2020-09-29
Trim whitespace

SuperSandro2000:


tulir:
The whitespace/newline is intentional, removing it would make it look like this:

image

SuperSandro2000:
But the diff shows it correct... but you are right. Sorry.

PR #6, opened 2020-10-06
Added ability to change direct chats name and avatar

witchent:
The "clean" method with "power_level_content_override" is not working, so this is the workaround. Looks cleaner that way anyway.

### DIFF
--- a/mautrix_signal/portal.py
+++ b/mautrix_signal/portal.py
@@ -634,7 +634,7 @@ class Portal(DBPortal, BasePortal):
         if self.is_direct:
             initial_state.append({
                 "type": "m.room.power_levels",
-                "content": {"users": {self.bridge_info.get('creator'): 100},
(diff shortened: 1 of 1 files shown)


mautrix_signal/portal.py
```diff
@@ -630,6 +630,13 @@ async def _create_matrix_room(self, source: 'u.User', info: ChatInfo) -> Optiona
                 "type": "m.room.related_groups",
                 "content": {"groups": [self.config["appservice.community_id"]]},
             })
+        #Allow chaning of room avatar and name in direct chats
+        if self.is_direct:
+            initial_state.append({
+                "type": "m.room.power_levels",
+                "content": {"users": {self.bridge_info.get('creator'): 100},
```
tulir, line 637:
```suggestion
                "content": {"users": {self.main_intent.mxid: 100},
```

PR #8, opened 2020-10-29
Make verify code work without signal bot

witchent:
Also made the phone number (hopefully) optional if in a portal

### DIFF
--- a/mautrix_signal/commands/signal.py
+++ b/mautrix_signal/commands/signal.py
@@ -81,7 +81,7 @@ async def pm(evt: CommandEvent) -> None:
 
 @command_handler(needs_auth=True, management_only=False, help_section=SECTION_SIGNAL,
                  help_text="View the safety number of a specific user",
-                 help_args="[--qr] <_phone_>")
+                 help_args="[--qr] [_phone_]")
 async def safety_number(evt: CommandEvent) -> None:
     show_qr = evt.args and evt.args[0].lower() == "--qr"
     if show_qr:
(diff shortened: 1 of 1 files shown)


mautrix_signal/commands/signal.py
```diff
@@ -81,15 +81,16 @@ async def pm(evt: CommandEvent) -> None:
 
 @command_handler(needs_auth=True, management_only=False, help_section=SECTION_SIGNAL,
                  help_text="View the safety number of a specific user",
-                 help_args="[--qr] [_phone_]")
+                 help_args="[--qr] <_phone_>")
```
tulir, line 84: This was the wrong way initially, so now it should actually be square brackets

```suggestion
                 help_args="[--qr] [_phone_]")
```

mautrix_signal/commands/signal.py
```diff
@@ -81,15 +81,16 @@ async def pm(evt: CommandEvent) -> None:
...
     show_qr = evt.args and evt.args[0].lower() == "--qr"
     if show_qr:
         if not qrcode:
             await evt.reply("Can't generate QR code: qrcode and/or PIL not installed")
             return
         evt.args = evt.args[1:]
-
+    if len(evt.args) == 0 and evt.is_portal:
+        evt.args = [evt.portal.chat_id.number]
     puppet = await _get_puppet_from_cmd(evt)
```
tulir, line 94:
```suggestion
    if len(evt.args) == 0 and evt.is_portal:
        puppet = await pu.Puppet.get_by_address(evt.portal.chat_id)
    else:
        puppet = await _get_puppet_from_cmd(evt)
```

witchent (commented):
Tulir's feedback

Issue #14, opened 2020-11-18
Additional configuration for "incoming" attachment directory (with finn/signald docker container)


tobymurray:
So far as I understand, the bridge/signal deployment is intended to share the attachment directory, so there's only ever one copy of the attachments. If that's incorrect, I imagine most of what follows will be as well:

I tried to use a docker-compose deployment making use of the finn/signald docker container ([Dockerfile](https://gitlab.com/thefinn93/signald/-/blob/master/Dockerfile) and [image](https://hub.docker.com/r/finn/signald)

```
services:
  mautrix-signal:
    container_name: mautrix-signal
    image: dock.mau.dev/tulir/mautrix-signal
    restart: unless-stopped
    volumes:
    - ./bridge:/data
    - ./signald:/signald
    depends_on:
    - signald
  signald:
    container_name: signald
    image: finn/signald
    restart: unless-stopped
    volumes: 
    - ./signald:/signald
```

One of the more significant differences between the two signald docker containers appears to be the directory where attachments are stored (`/signald/attachments` vs `/var/run/signald/attachments`). So far as I can tell, the mautrix-signal configuration for signald can only be used if the attachment directory has the same path in both the bridge container and the signald container. E.g. with the above docker-compose file and an `outgoing_attachment_dir: /signald/attachments`, that path is used when querying signald:
```
signald           | 14:42:15.163 [socketlistener] ERROR io.finn.signald.SocketHandler - Catching
signald           | org.asamk.signal.AttachmentInvalidException: /signald/attachments/mautrix-signal-ca29c1ae-fc0f-48a4-a0a7-64f3fe16587d: /signald/attachments/mautrix-signal-ca29c1ae-fc0f-48a4-a0a7-64f3fe16587d (No such file or directory)
signald           | 	at io.finn.signald.SocketHandler.send(SocketHandler.java:255) ~[signald.jar:unspecified]
signald           | 	at io.finn.signald.SocketHandler.handleRequest(SocketHandler.java:137) [signald.jar:unspecified]
signald           | 	at io.finn.signald.SocketHandler.run(SocketHandler.java:106) [signald.jar:unspecified]
signald           | 	at java.lang.Thread.run(Thread.java:748) [?:1.8.0_272]
mautrix-signal    | [2020-11-17 14:42:15,164] [ERROR@mau.as] Exception in Matrix event handler
mautrix-signal    | Traceback (most recent call last):
[...]

PR #15, opened 2020-11-22
Addresses #14: Add configuration for incoming attachment directory

tobymurray:
As far as I'm aware, this is exclusively useful in the context of a Docker deployment where signald and mautrix-signal are in different containers AND the paths to their respective data directories are not identical.

This introduces a new configuration item, `incoming_attachment_dir`. When set and receiving/sending attachments, this translates a path from being mautrix-relative to being signald-relative. From an implementation perspective, this is on the "mautrix" side of things (portal.py). It could conceivably be on the "signald" side of things (signald.py) instead I found it was easier to modify the path on creation.

This enables easier integration with the `finn/signald` Docker image. I've tried it out for only basic use cases (sending pictures from Matrix and receiving pictures from Signal), and only for the Docker deployment case.

### DIFF
--- a/mautrix_signal/example-config.yaml
+++ b/mautrix_signal/example-config.yaml
@@ -75,6 +75,11 @@ signal:
     # absolute path that signald can read. For attachments in the other direction,
     # make sure signald is configured to use an absolute path as the data directory.
     outgoing_attachment_dir: /tmp
+    # Directory for temp files when receiving files from Signal. For most deployments, this should
+    # not be specified (defaults to use outgoing_attachment_dir). If you're deploying signald and
+    # this bridge in two separate Docker containers, this is the absolute path to the data
+    # directory in signald.
+    # incoming_attachment_dir: /tmp
     # Directory where signald stores avatars for groups.
     avatar_dir: ~/.config/signald/avatars
     # Whether or not message attachments should be removed from disk after they're bridged.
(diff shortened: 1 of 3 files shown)


mautrix_signal/example-config.yaml
```diff
@@ -75,6 +75,11 @@ signal:
     # absolute path that signald can read. For attachments in the other direction,
     # make sure signald is configured to use an absolute path as the data directory.
     outgoing_attachment_dir: /tmp
+    # Directory for temp files when receiving files from Signal. For most deployments, this should
+    # not be specified (defaults to use outgoing_attachment_dir). If you're deploying signald and
+    # this bridge in two separate Docker containers, this is the absolute path to the data
+    # directory in signald.
+    # incoming_attachment_dir: /tmp
```
tobymurray, line 82: I left this commented out with the intention of not forcing it on everyone, but it appears to be the only commented out configuration key (as far as I noticed at least). That makes it a bit hinky...

tobymurray:
With https://gitlab.com/thefinn93/signald/-/merge_requests/34, the `finn/signald` image has been aligned with `dock.mau.dev/maunium/signald` in regards to data directory. The issue this PR addresses should no longer be a concern, though theoretically it's still valid (works by "coincidence" presently, relying on the same paths in two distinct filesystems).

PR #20, opened 2020-12-11
Automatically reconnect to signald

### LINKED ISSUE #18: Bridge breaks when signald restarts
If signald gets restarted while mautrix-signal is running, the bridge breaks until mautrix-signal is restarted.  Any messages sent from the Matrix while the bridge is broken seem to get lost forever.

```
[2020-12-02 16:19:37,171] [DEBUG@mausignald] Reader disconnected
[2020-12-02 16:19:52,389] [DEBUG@mau.mx] Received Matrix event <redacted>
[2020-12-02 16:19:52,402] [ERROR@mau.as] Exception in Matrix event handler
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/mautrix/appservice/as_handler.py", line 178, in try_handle
    await handler_func(event)
  File "/usr/lib/python3.8/site-packages/mautrix/bridge/matrix.py", line 441, in int_handle_event
    await self.handle_encrypted(evt)
  File "/usr/lib/python3.8/site-packages/mautrix/bridge/matrix.py", line 361, in handle_encrypted
    await self.int_handle_event(decrypted)
  File "/usr/lib/python3.8/site-packages/mautrix/bridge/matrix.py", line 439, in int_handle_event
    await self.handle_message(evt.room_id, evt.sender, evt.content, evt.event_id)
  File "/usr/lib/python3.8/site-packages/mautrix/bridge/matrix.py", line 282, in handle_message
    await portal.handle_matrix_message(sender, message, event_id)
  File "/usr/lib/python3.8/site-packages/mautrix_signal/portal.py", line 193, in handle_matrix_message
    await self.signal.send(username=sender.username, recipient=self.chat_id, body=text,
  File "/opt/mautrix-signal/mausignald/signald.py", line 134, in send
    await self.request("send", "send_results", username=username, messageBody=body,
  File "/opt/mautrix-signal/mausignald/rpc.py", line 157, in request
    resp_type, resp_data = await self._raw_request(command, **data)
  File "/opt/mautrix-signal/mausignald/rpc.py", line 153, in _raw_request
    await self._send_request(data)
  File "/opt/mautrix-signal/mausignald/rpc.py", line 145, in _send_request
    self._writer.write(json.dumps(data).encode("utf-8"))
AttributeError: 'NoneType' object has no attribute 'write'
```

It looks like `SignaldRPCClient` is missing reconnection logic when the read loop terminates.

satwell:
If the signald socket gets disconnected (e.g., signald restart), we will
now retry reconnecting to the socket forever, with a 5 second wait
between attempts.  We will also retry the initial connect forever if the
socket connection fails on startup.

This change creates a couple "synthetic" RPC events for connect and
disconnect that handlers can be attached to, as well as a NotConnected
exception class.  These should help callers deal with the fact that the
client may not always be connected.

I think creating these synthetic events that aren't triggered by messages from signald may be a little awkward.  But we need some way for library callers to register callbacks for reconnections, and using the existing handler code in rpc.py seems like the cleanest solution to me.  Happy to re-work this if you prefer something else.

I've done a little bit of manual testing, and this seems to work well.  We still lose any messages sent from matrix while disconnected.  But it should now be easy to buffer these failed messages and send after reconnect.

Fixes #18

### DIFF
--- a/mausignald/errors.py
+++ b/mausignald/errors.py
@@ -27,6 +27,10 @@ class LinkingError(RPCError):
         self.number = number
 
 
+class NotConnected(RPCError):
+    pass
+
+
 class LinkingTimeout(LinkingError):
     pass
 
--- a/mausignald/rpc.py
+++ b/mausignald/rpc.py
@@ -11,9 +11,14 @@ import json
 
 from mautrix.util.logging import TraceLogger
 
-from .errors import UnexpectedError, UnexpectedResponse
+from .errors import NotConnected, UnexpectedError, UnexpectedResponse
 
 EventHandler = Callable[[Dict[str, Any]], Awaitable[None]]
+
+# These are synthetic RPC events for registering callbacks on socket
+# connect and disconnect.
+CONNECT_EVENT = "_socket_connected"
+DISCONNECT_EVENT = "_socket_disconnected"
 
 
 class SignaldRPCClient:
@@ -23,6 +28,7 @@ class SignaldRPCClient:
     socket_path: str
     _reader: Optional[asyncio.StreamReader]
     _writer: Optional[asyncio.StreamWriter]
+    _communicate_task: Optional[asyncio.Task]
 
     _response_waiters: Dict[UUID, asyncio.Future]
     _rpc_event_handlers: Dict[str, List[EventHandler]]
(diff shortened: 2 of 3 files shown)


tulir:
Looks fine, can you rebase it to master?

satwell:
Rebased to master.

PR #23, opened 2020-12-18
Add groupname_template to change naming of Signal groups

LarsHaalck:
In order to distinguish Signal and Matrix rooms, I found it handy to append "(Signal)" as it is the default for display names in 1:1 chats. I only added another variable in `config.yaml` and changed the `update_info()`-function for `Group` and `GroupV2` instances.

### DIFF
--- a/mautrix_signal/example-config.yaml
+++ b/mautrix_signal/example-config.yaml
@@ -79,6 +79,9 @@ bridge:
     # available variable in displayname_preference. The variables in displayname_preference
     # can also be used here directly.
     displayname_template: "{displayname} (Signal)"
+    # Groupname template for Signal groups.
+    # {} is replaced with the name of the group
+    groupname_template: "{} (Signal)"
     # Whether or not contact list displaynames should be used.
     # Using this isn't recommended on multi-user instances.
     allow_contact_list_name_updates: false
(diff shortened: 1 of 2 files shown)


steadfasterX:
nice idea! I would appreciate having that :)

maltee1:
Just in case this PR is still being considered, my 2 cents:
This will probably cause some undesirable behavior in a few scenarios that I can come up with:
When renaming the group on matrix, the existing "(Signal)" suffix will be sent to signal, which will on the next sync send a group name back with that suffix, which the bridge will turn into "(Signal) (Signal)".
Even worse when this is implemented on several bridges that support relaying room name changes and are used in the same room. It will cause a never-ending ping-pong of renaming the group - until the maximum group name length is reached, that is.
So this would have to be done a bit smarter, maybe the bridge should always scrap (Signal) from the group name before sending it to signal, no matter where in the name it reads it, or simply not add it when bridging the name from signal to matrix and (Signal) is already there. And it would probably have to be disabled if any non-logged-in users are in the room.

I think it's best avoided, because it's tricky to get right.

LarsHaalck:
Personally, I don't care about this PR anymore since I don't use the bridge due to multiple issues unrelated to this.

If someone is still interested, feel free to use this PR as a basis for any further changes.

PR #113, opened 2021-08-16
send: add m.notice when send to Signal fails

sumnerevans:


mautrix_signal/portal.py
```diff
@@ -286,9 +286,17 @@ async def handle_matrix_message(self, sender: 'u.User', message: MessageEventCon
...
+        try:
+            await self.signal.send(username=sender.username, recipient=self.chat_id, body=text,
+                                   mentions=mentions, quote=quote, attachments=attachments,
+                                   timestamp=request_id)
+        except Exception as e:
+            await self.main_intent.send_notice(
+                self.mxid,
+                f"\u26a0 Your message was not bridged: {e}",
+                html="\u26a0 Your message was not bridged:<br>{}".format(
+                    str(e).replace("\n", "<br>")))
```
tulir, line 298: Does this need HTML? Seems like it doesn't actually have any formatting, newlines work fine in plaintext. It should also use `._send_message()` so the message would get encrypted. Something like this:

```suggestion
            await self._send_message(
                self.main_intent,
                TextMessageEventContent(
                    msgtype=MessageType.NOTICE,
                    body=f"\u26a0 Your message was not bridged: {e}"))
```

sumnerevans: hmm... for some reason it seemed like it wasn't respecting newlines, but I may have been some other issue that I caused.

tulir (approved):

PR #153, opened 2021-11-01
Make it possible to disable device registration from the config

tadzik:
Handy for running a linked-devices-only bridge, which Signal treats a lot more liberally.

### DIFF
--- a/mautrix_signal/commands/auth.py
+++ b/mautrix_signal/commands/auth.py
@@ -81,6 +81,7 @@ async def link(evt: CommandEvent) -> None:
 
 
 @command_handler(needs_auth=False, management_only=True, help_section=SECTION_AUTH,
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -4,5 +4,5 @@ commonmark>=0.8,<0.10
 aiohttp>=3,<4
 yarl>=1,<2
 attrs>=19.1
-mautrix>=0.10.5,<0.11
+mautrix>=0.10.10,<0.11
```
tulir, line 7: This caused a merge conflict since the dependency was already bumped to 0.11+

tulir (approved):

requirements.txt
tadzik: Fixed now with a rebase.

PR #154, opened 2021-11-02
send: handle proof_required_failure

sumnerevans:


tulir (approved):
We should probably deserialize the send response properly at some point :thinking:

PR #163, opened 2021-11-09
Bridge metrics and reconnect

Half-Shot:
This change does two things:
- Adds a bunch of metrics around problem paths we've found. Notably disconnections and profile requests
- Attempts to reconnect on disconnect, I don't know how successful this will be but worth a go?


Half-Shot:
Sigh, wish I could remember the order of these changes.

PR #164, opened 2021-11-09
Add additional metrics to the bridge

Half-Shot:
This PR adds a few choice metrics to the bridge to handle things which we're currently trying to monitor for health reasons.

In addition, there are a few places where the bridge can either fallover on startup or silently fail and it was useful to add logging to those places.

### DIFF
--- a/mausignald/rpc.py
+++ b/mausignald/rpc.py
@@ -73,11 +73,11 @@ class SignaldRPCClient:
                 self._reader, self._writer = await asyncio.open_unix_connection(
                     self.socket_path, limit=_SOCKET_LIMIT)
             except OSError as e:
-                self.log.error(f"Connection to {self.socket_path} failed: {e}")
+                self.log.error(f"Connection to {self.socket_path} failed: {e}, retrying in 5s")
                 await asyncio.sleep(5)
                 continue
-            self.log.info(f"Connection to {self.socket_path} succeeded")
 
+            self.log.info(f"Connection to {self.socket_path} succeeded")
             read_loop = asyncio.create_task(self._try_read_loop())
             self.is_connected = True
             CONNECTED_GAUGE.set(1)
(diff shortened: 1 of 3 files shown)


mautrix_signal/signal.py
```diff
@@ -149,12 +149,17 @@ async def handle_receipt(sender: 'pu.Puppet', receipt: Receipt) -> None:
...
-            known_usernames.add(user.username)
-            if await self.subscribe(user.username):
-                asyncio.create_task(user.sync())
+            try:
+                known_usernames.add(user.username)
+                if await self.subscribe(user.username):
+                    asyncio.create_task(user.sync())
+            except Exception as ex:
+                # Don't hold up the whole bridge over this.
+                self.log.warn(f"Failed to subscribe to {user.username}: {ex}")
```
tulir, line 162: https://docs.python.org/3/library/logging.html#logging.Logger.warning

```suggestion
                self.log.warning(f"Failed to subscribe to {user.username}: {ex}")
```

mausignald/rpc.py
```diff
@@ -59,29 +63,37 @@ async def wait_for_connected(self, timeout: Optional[int] = None) -> bool:
...
+                    CONNECTED_GAUGE.set(1)
+                    await self._run_rpc_handler(CONNECT_EVENT, {})
+                    self._connect_future.set_result(True)
+
+                    await read_loop
+                    self.is_connected = False
+                    CONNECTED_GAUGE.set(0)
+                    await self._run_rpc_handler(DISCONNECT_EVENT, {})
+                    self._connect_future = self.loop.create_future()
+                    RECONNECTIONS_COUNTER.inc(1)
```
tulir, line 93: This block seems over-indented

Half-Shot: Have split out the loop from the code, i'd still like the try catch to just log when we're going to have a fatal error, because otherwise this failed silently (when we have something catastrophic)

tulir: I meant it was indented with 4 spaces more than needed, but that works too

Half-Shot: Oh, I thought you reasonably objected to massive try catch blocks :D. well, hopefully it's less

PR #166, opened 2021-11-09
Support mautrix@0.11.0

Half-Shot:
From testing, this seemed to be the only thing that broke in the release.


sumnerevans (approved):

Half-Shot:
Hah, so just a bit more work than I thought then :laughing: . Thanks!

tulir:
Well most of it was for SQLite support (which is technically not required to update), but the `.upgrade` and `.override_pool` parameters definitely needed to be updated

PR #169, opened 2021-11-10
Add some jitter to the sync loop

### LINKED ISSUE #168: Syncing of profile information (and contact info) needs to be staggered to prevent thundering herd
We've added metrics to track the results of profile requests in the bridge and noticed that it has been doing this:

image

Which feels a bit bleugh, perhaps there should be some noise introduced so that requests are handled roughly every N seconds rather than exactly.

Half-Shot:
Fixes #168 

This should ensure that sync requests made to signald are appropriately staggered to prevent the process from being hit by a thundering herd of requests. At the moment this is hardcoded to 10s.

There is a risk that this will make the profile sync speed lower than advertised in the config, but overall shouldn't be too noticeable for smaller bridges.

### DIFF
--- a/mautrix_signal/__main__.py
+++ b/mautrix_signal/__main__.py
@@ -14,6 +14,7 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 from typing import Dict, Any
+from random import uniform
 import asyncio
 import logging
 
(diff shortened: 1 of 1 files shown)


mautrix_signal/__main__.py
```diff
@@ -33,6 +34,7 @@
 from .web import ProvisioningAPI
 from . import commands
 
+SYNC_JITTER=10
```
sumnerevans, line 37: super nit:
```suggestion
SYNC_JITTER = 10
```

Half-Shot: fair nit!

sumnerevans (approved):
seems reasonable to me

tulir (approved):

PR #182, opened 2021-12-09
disappearing messages: add support for both inbound and outbound

sumnerevans:
When a message is sent or recieved in a Signal chat that has disappearing messages enabled, it will automatically be redacted by the bridge after the time configured on the room. However, note that the countdown timer will only start once the room has been read.

The main mechanism for this is using async functions that just wait the configured number of seconds before redacting. However, there we also store all of the state necessary for determining when to redact a message in the database in case a restart occurs. In the event of a restart, we can resume waiting for the expiration (or redact immediately if we are past the expiration timestamp).

This change also adds m.notice notifications for the user when the disappearing messages setting changes on the Signal chat.

### DIFF
--- a/mautrix_signal/db/disappearing_message.py
+++ b/mautrix_signal/db/disappearing_message.py
@@ -59,10 +59,6 @@ class DisappearingMessage:
         await cls.db.execute(q, room_id, event_id)
 
     @classmethod
-    async def delete_all(cls, room_id: RoomID) -> None:
-        await cls.db.execute("DELETE FROM message WHERE room_id=$1", room_id)
-
-    @classmethod
     def _from_row(cls, row: asyncpg.Record) -> "DisappearingMessage":
         return cls(**row)
 
--- a/mautrix_signal/db/disappearing_message.py
+++ b/mautrix_signal/db/disappearing_message.py
@@ -1,0 +1,94 @@
+# mautrix-signal - A Matrix-Signal puppeting bridge
+# Copyright (C) 2021 Sumner Evans
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+from typing import ClassVar, List, Optional, TYPE_CHECKING
+
+from attr import dataclass
+import asyncpg
+
+from mautrix.types import RoomID, EventID
+from mautrix.util.async_db import Database
+
+fake_db = Database.create("") if TYPE_CHECKING else None
+
+
+@dataclass
+class DisappearingMessage:
+    db: ClassVar[Database] = fake_db
+
+    room_id: RoomID
+    mxid: EventID
+    expiration_seconds: int
+    expiration_ts: Optional[int] = None
+
+    async def insert(self) -> None:
+        q = """
+        INSERT INTO disappearing_messages (room_id, mxid, expiration_seconds, expiration_ts)
+        VALUES ($1, $2, $3, $4)
+        """
+        await self.db.execute(q, self.room_id, self.mxid, self.expiration_seconds,
+                              self.expiration_ts)
+
+    async def update(self) -> None:
+        q = """
+        UPDATE disappearing_messages
+        SET expiration_seconds=$3, expiration_ts=$4
+        WHERE room_id=$1 AND mxid=$2
+        """
+        try:
+            await self.db.execute(q, self.room_id, self.mxid, self.expiration_seconds,
+                              self.expiration_ts)
+        except Exception as e:
+            print(e)
+
+    @classmethod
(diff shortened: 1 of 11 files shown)


mautrix_signal/portal.py
```diff
@@ -150,6 +156,10 @@ def init_cls(cls, bridge: 'SignalBridge') -> None:
         BasePortal.bridge = bridge
         cls.private_chat_portal_meta = cls.config["bridge.private_chat_portal_meta"]
 
+        for dm in await DisappearingMessage.get_all():
+            if dm.expiration_ts:
+                asyncio.create_task(cls._expire_event(dm))
```
tulir, line 161: This should probably be in a separate method that's run after startup, `init_cls` shouldn't be async

mautrix_signal/portal.py
```diff
@@ -328,6 +338,14 @@ async def handle_matrix_message(self, sender: 'u.User', message: MessageEventCon
                 except FileNotFoundError:
                     pass
 
+            # Handle disappearing messages
+            if self.expiration_time:
+                disappearing_message = DisappearingMessage(
+                    self.mxid, event_id, self.expiration_time
+                )
+                await disappearing_message.insert()
+                asyncio.create_task(Portal._expire_event(disappearing_message))
```
tulir, line 347:
```suggestion
                asyncio.create_task(self._expire_event(disappearing_message))
```

mautrix_signal/portal.py
```diff
@@ -522,6 +540,59 @@ async def handle_matrix_avatar(self, user: 'u.User', url: ContentURI) -> None:
...
+        mechanism to stop the countdown, even after bridge restart.
+        """
+        room_id = disappearing_message.room_id
+        event_id = disappearing_message.mxid
+        wait = disappearing_message.expiration_seconds
+        # If there is an expiration_ts, then there was probably a bridge restart, so we have to
+        # resume the countdown. This is fairly likely to occur if the disappearance timeout is
+        # weeks.
+        # If there is not an expiration_ts, then set one.
+        now = datetime.now().timestamp()
```
tulir, line 557:
```suggestion
        now = time.time()
```

tulir (changes requested):
This isn't going to work for groups on multi-user instances, so it should probably be a config option

mautrix_signal/portal.py
sumnerevans: I went with `Portal.` since it is a class method.

mautrix_signal/db/disappearing_message.py
```diff
@@ -0,0 +1,94 @@
...
+            print(e)
+
+    @classmethod
+    async def delete(cls, room_id: RoomID, event_id: EventID) -> None:
+        q = "DELETE from disappearing_messages WHERE room_id=$1 AND mxid=$2"
+        await cls.db.execute(q, room_id, event_id)
+
+    @classmethod
+    async def delete_all(cls, room_id: RoomID) -> None:
+        await cls.db.execute("DELETE FROM message WHERE room_id=$1", room_id)
```
tulir, line 63: It probably shouldn't delete all messages here :D (are we using this at all?)

mautrix_signal/db/upgrade.py
```diff
@@ -175,3 +175,16 @@ async def upgrade_v6(conn: Connection) -> None:
 @upgrade_table.register(description="Add relay user field to portal table")
 async def upgrade_v7(conn: Connection) -> None:
     await conn.execute("ALTER TABLE portal ADD COLUMN relay_user_id TEXT")
+
+
+@upgrade_table.register(description="Add support for disappearing messages")
+async def upgrade_v8(conn: Connection) -> None:
+    await conn.execute("""CREATE TABLE disappearing_messages (
```
tulir, line 182: Table name should be singular for consistency

```suggestion
    await conn.execute("""CREATE TABLE disappearing_message (
```

mautrix_signal/portal.py
```diff
@@ -522,6 +542,59 @@ async def handle_matrix_avatar(self, user: 'u.User', url: ContentURI) -> None:
...
+        else:
+            disappearing_message.expiration_ts = int((now + wait) * 1000)
+            await disappearing_message.update()
+
+        if wait < 0:
+            wait = 0
+
+        portal = await cls.get_by_mxid(room_id)
+        if not portal:
+            portal.log.warning(f"No portal for {room_id}")
```
tulir, line 571: This looks a bit `AttributeError`y (and none of the rest of the function will work without the `portal` either)

sumnerevans: Fixed. I meant to return early there, but clearly forgot. Updated to throw an exception and moved the check earlier.

mautrix_signal/portal.py
```diff
@@ -522,6 +548,71 @@ async def handle_matrix_avatar(self, user: 'u.User', url: ContentURI) -> None:
...
+            except Exception as e:
+                portal.log.warning("Redacting expired event didn't work", e)
+            finally:
+                await DisappearingMessage.delete(room_id, event_id)
+
+    async def handle_read_receipt(self, event_id: EventID, data: SingleReceiptEventContent):
+        # Start the redaction timers for all of the disappearing messages in the room when the user
+        # reads the room. This is the behavior of the Signal clients.
+        for disappearing_message in await DisappearingMessage.get_all_for_room(self.mxid):
+            asyncio.create_task(Portal._expire_event(disappearing_message))
```
tulir, line 614: Won't this create a new task for each message in the room whenever there's a read receipt?

sumnerevans: oh, good point. Let me fix...

tulir (approved):
There's definitely room for optimization but probably fine as is for the initial version

PR #184, opened 2021-12-09
calls: show notifications for call start/end in DMs

sumnerevans:
image

### DIFF
--- a/mautrix_signal/signal.py
+++ b/mautrix_signal/signal.py
@@ -70,6 +70,8 @@ class SignalHandler(SignaldClient):
             await self.handle_typing(user, sender, evt.typing)
         if evt.receipt:
             await self.handle_receipt(sender, evt.receipt)
+        if evt.call_message:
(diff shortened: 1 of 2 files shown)


mautrix_signal/signal.py
```diff
@@ -116,6 +118,38 @@ async def handle_message(self, user: 'u.User', sender: 'pu.Puppet', msg: Message
...
+        if msg.call_message.offer_message:
+            call_type = {
+                OfferMessageType.AUDIO_CALL: " voice ",
+                OfferMessageType.VIDEO_CALL: " video ",
+            }.get(msg.call_message.offer_message.type, " ")
+            msg_html += (
+                f" started a{call_type}call on Signal. Use the native app to answer the call."
+            )
+        elif msg.call_message.hangup_message:
+            msg_html += " ended a call on Signal."
```
tulir, line 145: Call end notifications should be disabled by default and/or use `m.notice` instead of `m.text`

tulir (approved):

PR #198, opened 2022-01-18
reactions: add retry logic for failed reactions

sumnerevans:


mautrix_signal/portal.py
```diff
@@ -420,14 +422,20 @@ async def _handle_matrix_message(
                 await Portal._expire_event(dm.room_id, dm.mxid)
 
     async def _signal_send_with_retries(
-        self, sender: u.User, message: MessageEventContent, event_id: EventID, **send_args
+        self,
+        sender: u.User,
+        event_id: EventID,
+        send_fn: Callable,
+        event_type: EventType,
+        message_type: MessageEventContent | None = None,
```
tulir, line 430:
```suggestion
        message_type: MessageType | None = None,
```

mautrix_signal/portal.py
```diff
@@ -476,11 +484,41 @@ async def handle_matrix_reaction(
...
+            await self._send_delivery_receipt(event_id)
 
+    async def _handle_matrix_reaction(
+        self, sender: u.User, event_id: EventID, reacting_to: EventID, emoji: str
+    ) -> None:
         message = await DBMessage.get_by_mxid(reacting_to, self.mxid)
         if not message:
             self.log.debug(f"Ignoring reaction to unknown event {reacting_to}")
-            return
+            raise NotConnected(f"Ignoring reaction to unknown event {reacting_to}")
```
tulir, line 521: This looks like the catch in `_signal_send_with_retries` is going to log an unnecessary exception and send the checkpoint with a message that isn't true ("not connected to signald").

We should probably stop abusing existing exception classes and just make new ones when appropriate, a `class UnknownReactionTarget(Exception): pass` or something at the top of the file is fine. The handler can then catch both and adjust the log level and checkpoint message appropriately

tulir (changes requested):

mautrix_signal/portal.py
sumnerevans: OK, that sounds good. I wasn't really loving this implementation, so I think that adding a new exception class makes sense here.

tulir (approved):

PR #199, opened 2022-01-19
login: use wait_for_scan to figure out when the user scans the QR code

sumnerevans:
Requires https://gitlab.com/beeper/signald/-/merge_requests/12.

### DIFF
--- a/mautrix_signal/web/provisioning_api.py
+++ b/mautrix_signal/web/provisioning_api.py
@@ -25,7 +25,7 @@ from aiohttp import web
 from mausignald.errors import InternalError, TimeoutException
 from mausignald.types import Account, Address
 from mautrix.types import UserID
-from mautrix.util.bridge_state import BridgeStateEvent
+from mautrix.util.bridge_state import BridgeState, BridgeStateEvent
 from mautrix.util.logging import TraceLogger
 
 from .. import user as u
@@ -186,6 +186,18 @@ class ProvisioningAPI:
             )
         session_id = user.command_status["session_id"]
         device_name = user.command_status["device_name"]
+        try:
+            await self.bridge.signal.wait_for_scan(session_id)
+            status_endpoint = self.bridge.config["homeserver.status_endpoint"]
+            if status_endpoint:
+                state = BridgeState(state_event=BridgeStateEvent.CONNECTING).fill()
+                asyncio.create_task(state.send(status_endpoint, self.bridge.az.as_token, self.log))
+        except Exception as e:
+            self.log.exception(f"Failed waiting for scan. Error: {e}")
+            raise web.HTTPBadRequest(
+                text='{"error": "Failed to wait for scan"}', headers=self._headers
+            )
+
         try:
             account = await asyncio.shield(self._shielded_link(user, session_id, device_name))
         except asyncio.CancelledError:
--- a/mautrix_signal/web/provisioning_api.py
+++ b/mautrix_signal/web/provisioning_api.py
@@ -172,11 +172,10 @@ class ProvisioningAPI:
         try:
             account = await asyncio.shield(self._shielded_link(user, session_id, device_name))
         except asyncio.CancelledError:
-            error_text = f"Client cancelled link wait request ({session_id}) before it finished"
-            self.log.warning(error_text)
-            raise web.HTTPInternalServerError(
-                text=f'{{"error": "{error_text}"}}', headers=self._headers
+            self.log.warning(
+                f"Client cancelled link wait request ({session_id}) before it finished"
             )
+            raise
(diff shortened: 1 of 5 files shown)


mautrix_signal/web/provisioning_api.py
```diff
@@ -43,17 +43,27 @@ def __init__(self, bridge: "SignalBridge", shared_secret: str) -> None:
...
-        # self.app.router.add_post("/api/register/code", self.register_code)
-        self.app.router.add_post("/api/logout", self.logout)
+
+        # New Login API
+        self.app.router.add_options("/api/link/v2/new", self.login_options)
+        self.app.router.add_options("/api/link/v2/waitForScan", self.login_options)
+        self.app.router.add_options("/api/link/v2/waitForAccount", self.login_options)
+        self.app.router.add_post("/api/link/v2/new", self.link_new)
+        self.app.router.add_post("/api/link/v2/waitForScan", self.link_wait_for_scan)
+        self.app.router.add_post("/api/link/v2/waitForAccount", self.link_wait_for_account)
```
tulir, line 66: `/_matrix/provision/v1/api/link/v2/...` :upside_down_face: 

I'll try to figure out the best way to make the endpoint names less messy

tulir: Right, https://github.com/mautrix/signal/pull/199/commits/038fc52be38e27f9886ac27709255c61640e33b4 should make versioning sane. Legacy endpoints are unchanged, new endpoints are `/_matrix/provision/v1/link/wait/scan` (and `/new` and `/wait/account`).

There are v2 versions of whoami and logout to remove the redundant `/api` prefix. New versions can also be added easily if ever needed since the version is now in the endpoint path rather than global config file.

sumnerevans: I think that makes sense. I had actually thought about using `/link/wait/scan` and `/link/wait/account` as well, but I was 50/50 on which to do. I opted for a bit more explicit `v2`, but it makes a lot more sense having the `v2` before `/link`.

mautrix_signal/web/provisioning_api.py
```diff
@@ -21,11 +21,12 @@
 import logging
 
 from aiohttp import web
+from attr import asdict
```
tulir, line 24:
```suggestion
```

mautrix_signal/web/provisioning_api.py
```diff
@@ -212,6 +198,122 @@ async def link_wait(self, request: web.Request) -> web.Response:
...
+            )
+        session_id = user.command_status["session_id"]
+        device_name = user.command_status["device_name"]
+        return await self._try_shielded_link(user, session_id, device_name)
+
+    # endregion
+
+    # region New Link API
+
+    async def _get_request_data(self, request: web.Request) -> tuple[u.User, web.Response]:
```
tulir, line 242: This seems to return a dict (`await request.json()`)

```suggestion
    async def _get_request_data(self, request: web.Request) -> tuple[u.User, dict]:
```

mautrix_signal/web/provisioning_api.py
```diff
@@ -135,62 +149,34 @@ async def status(self, request: web.Request) -> web.Response:
...
         try:
             account = await asyncio.shield(self._shielded_link(user, session_id, device_name))
         except asyncio.CancelledError:
-            self.log.warning(
-                f"Client cancelled link wait request ({session_id}) before it finished"
+            error_text = f"Client cancelled link wait request ({session_id}) before it finished"
+            self.log.warning(error_text)
+            raise web.HTTPInternalServerError(
+                text=f'{{"error": "{error_text}"}}', headers=self._headers
             )
```
tulir, line 180: Raising a HTTP error here is pointless, the client is already gone. Might make sense to re-raise the CancelledError after logging it though

```suggestion
            self.log.warning(
                f"Client cancelled link wait request ({session_id}) before it finished"
            )
            raise
```

mautrix_signal/web/provisioning_api.py
```diff
@@ -212,6 +198,122 @@ async def link_wait(self, request: web.Request) -> web.Response:
...
+        except KeyError:
+            error_text = '{"error": "session_id not provided"}'
+            raise web.HTTPBadRequest(text=error_text, headers=self._headers)
+
+        try:
+            await self.bridge.signal.wait_for_scan(session_id)
+        except Exception as e:
+            error_text = f"Failed waiting for scan. Error: {e}"
+            self.log.exception(error_text)
+            self.log.info(e.__class__)
```
tulir, line 291: Leftover debug log?

```suggestion
```

tulir (approved):

PR #200, opened 2022-01-21
pre-commit: add configuration

sumnerevans:


tulir:
Apparently pre-commit/black is incompatible with `required-version`: https://github.com/psf/black/issues/2493

tulir (approved):

PR #206, opened 2022-01-28
rpc: add logging to correlate Matrix event ID to signald request ID

sumnerevans:

### DIFF
--- a/mautrix_signal/portal.py
+++ b/mautrix_signal/portal.py
@@ -438,8 +438,11 @@ class Portal(DBPortal, BasePortal):
         retry_message_event_id = None
         for retry_num in range(retry_count):
             try:
-                self.log.info(f"Send attempt {retry_num}")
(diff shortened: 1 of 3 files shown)


mautrix_signal/portal.py
```diff
@@ -386,6 +390,7 @@ async def _handle_matrix_message(
                 quote=quote,
                 attachments=attachments,
                 timestamp=request_id,
+                req_id=signald_req_id,
```
tulir, line 393: This looks like it'll use the same request ID for each retry, which feels wrong, but I guess it's fine if the retries are only for signald going down and/or if signald doesn't care about request ID uniqueness

tulir (approved):

mautrix_signal/portal.py
sumnerevans: Oh, good point.

PR #213, opened 2022-02-04
handle authorization failures in more places

sumnerevans:
This commit adds a utility function for helping to deal with auth failures and uses it around the codebase.

### DIFF
--- a/mautrix_signal/user.py
+++ b/mautrix_signal/user.py
@@ -21,6 +21,7 @@ from datetime import datetime
 from uuid import UUID
 import asyncio
 
+from mausignald.errors import ResponseError
 from mausignald.types import (
     Account,
     Address,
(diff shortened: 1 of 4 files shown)


mautrix_signal/user.py
```diff
@@ -309,25 +316,33 @@ async def _sync_group_v2(self, group: GroupV2, create_portals: bool) -> None:
...
-            except Exception:
-                self.log.exception(f"Failed to sync contact {contact.address}")
+        try:
+            for contact in await self.bridge.signal.list_contacts(self.username):
+                try:
+                    await self.sync_contact(contact, create_contact_portal)
+                except Exception:
+                    self.log.exception(f"Failed to sync contact {contact.address}")
+        except Exception as e:
+            await handle_auth_failure(self, e)
```
tulir, line 326: These catches should be in `sync_groups` and `sync_contacts` that are already just catch wrappers

mautrix_signal/util/handle_auth_failure.py
```diff
@@ -0,0 +1,27 @@
...
+from mautrix.util.bridge_state import BridgeStateEvent
+from mautrix_signal.user import User
+
+
+async def handle_auth_failure(user: User, e: Exception) -> None:
+    auth_failed = (
+        "org.whispersystems.signalservice.api.push.exceptions.AuthorizationFailedException"
+    )
+    if isinstance(e, ResponseError) and auth_failed in e.data.get("exceptions", []):
+        await user.push_bridge_state(BridgeStateEvent.BAD_CREDENTIALS, error=str(e))
```
tulir, line 27: This seems like it should just be a method in `User`

tulir (approved):

PR #222, opened 2022-02-14
Stop parsing InternalErrors

thefinn93:
signald's InternalError exception is meant to be a fallback when unexpected behavior occurs. We were parsing it out in several places which have now been converted into proper protocol errors on the signald side

### DIFF
--- a/mausignald/errors.py
+++ b/mausignald/errors.py
@@ -53,7 +53,11 @@ class CaptchaRequired(ResponseError):
     pass
 
 
-class AuthorizationFailedException(ResponseError):
+class AuthorizationFailedError(ResponseError):
+    pass
+
+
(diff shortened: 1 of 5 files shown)


mausignald/errors.py
```diff
@@ -53,7 +53,11 @@ class CaptchaRequired(ResponseError):
     pass
 
 
-class AuthorizationFailedException(ResponseError):
+class AuthorizationFailedError(ResponseError):
+    pass
+
+
+class ScanTimeoutError(ResponseError):
```
tulir, line 60: This probably needs to be added to `response_error_types` too

thefinn93: done

tulir (approved):

PR #223, opened 2022-02-14
signal kicks: handle being kicked from Signal group

sumnerevans:
image

### DIFF
--- a/mautrix_signal/portal.py
+++ b/mautrix_signal/portal.py
@@ -945,6 +945,10 @@ class Portal(DBPortal, BasePortal):
         else:
             self.log.debug(f"Didn't get event ID for {message.timestamp}")
 
+    async def handle_signal_kicked(self, user: u.User, sender: pu.Puppet) -> None:
+        self.log.debug(f"{user.mxid} was kicked by {sender.number} from {self.mxid}")
+        await self.main_intent.kick_user(self.mxid, user.mxid, f"{sender.name} kicked you")
+
     @staticmethod
     async def _make_media_content(
         attachment: Attachment, data: bytes
--- a/mautrix_signal/signal.py
+++ b/mautrix_signal/signal.py
@@ -175,6 +175,13 @@ class SignalHandler(SignaldClient):
                 )
                 return
         assert portal
+
+        # Handle the user being removed from the group.
+        if msg.group_v2 and msg.group_v2.removed:
+            if portal.mxid:
+                await portal.handle_signal_kicked(user, sender)
+            return
+
         if not portal.mxid:
             if not msg.is_message and not msg.group_v2:
                 user.log.debug(
--- a/mausignald/types.py
+++ b/mausignald/types.py
@@ -163,6 +163,7 @@ class Group(SerializableAttrs):
 class GroupV2ID(SerializableAttrs):
     id: GroupID
     revision: Optional[int] = None
+    removed: Optional[bool] = False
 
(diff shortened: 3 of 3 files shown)


mautrix_signal/portal.py
```diff
@@ -945,6 +945,13 @@ async def handle_signal_message(
         else:
             self.log.debug(f"Didn't get event ID for {message.timestamp}")
 
+    async def handle_signal_kicked(self, user: u.User, sender: pu.Puppet) -> None:
+        self.log.debug(f"{user.mxid} was kicked by {sender.number} from {self.mxid}")
+        await sender.intent_for(self).send_notice(self.mxid, "Removed you from the chat.")
+        levels = await self.main_intent.get_power_levels(self.mxid)
+        levels.events_default = 50
+        await self.main_intent.set_power_levels(self.mxid, levels)
```
tulir, line 953: This doesn't work for multi-user bridges, it should probably just actually kick the user

mautrix_signal/signal.py
```diff
@@ -162,6 +163,13 @@ async def handle_message(
             return
         if msg.group_v2:
             portal = await po.Portal.get_by_chat_id(msg.group_v2.id, create=True)
+
+            try:
+                await self.get_group(user.username, msg.group_v2.id)
+            except UnknownGroupError as e:
+                # The user was removed from the group. Make the user unable to post to the room.
+                await portal.handle_signal_kicked(user, sender)
+                return
```
tulir, line 172: This is a bit suspicious since historically there have been lots of `UnknownGroupError`s (e.g. when signald hadn't synced groups), but I guess it might be fine if it's only caught when the message explicitly contains group metadata and if there's no other way to detect being removed

thefinn93: what other scenario does the bridge know about a group and show it to the user, but signald doesn't know about it?

tulir: Probably just logout/relogins (signald data is removed, bridge data is still there). How reliable is the group data syncing now?

thefinn93: ah that is true. I don't have metrics about storage service usage, but it Works For Me every time.

maybe the message about being removed from the group could indicate there are edge cases around re-linking and to send a message in this group from the linked device to make it visible in matrix

tulir: The signald store and signal event handling is separate for different users even if it's the same bridge, so I think it's fine. Would definitely prefer a clear "kicked from group" event though.

sumnerevans: @tulir do we actually care about situations where signald's database got broken but mautrix-signal is fine? Can you actually do that in Beeper these days?

thefinn93: Working on getting you an explicit `removed: true` `JsonGroupV2Info` in inbound messages

mautrix_signal/signal.py
```diff
@@ -175,6 +175,12 @@ async def handle_message(
                 )
                 return
         assert portal
+
+        # Handle the user being removed from the group.
+        if msg.group_v2 and msg.group_v2.removed:
+            await portal.handle_signal_kicked(user, sender)
```
tulir, line 181: The room isn't guaranteed to exist here

```suggestion
            if portal.mxid:
                await portal.handle_signal_kicked(user, sender)
```

tulir (approved):

PR #236, opened 2022-03-31
fix docker user

thefinn93:
associated signald change incoming


tulir:
This needs to update the log config in `fixperms` and also call the function after generating a registration/config (see https://github.com/mautrix/instagram/blob/master/docker-run.sh for example)

tulir:
Just needs https://github.com/mautrix/instagram/blob/master/docker-run.sh#L22 and https://github.com/mautrix/instagram/blob/master/docker-run.sh#L31

tulir (approved):

thefinn93:
Gonna leave this open until the signald side gets sorted out

PR #245, opened 2022-04-15
Leave group

maltee1:
This seems to work. Please take a look.
But I don't really know python, I mostly used copy & paste and common sense. I also don't know what possible issues to test for.

### DIFF
--- a/mautrix_signal/example-config.yaml
+++ b/mautrix_signal/example-config.yaml
@@ -194,6 +194,8 @@ bridge:
     resend_bridge_info: false
     # Interval at which to resync contacts (in seconds).
     periodic_sync: 0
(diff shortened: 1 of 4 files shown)


mautrix_signal/example-config.yaml
```diff
@@ -194,6 +194,8 @@ bridge:
     resend_bridge_info: false
     # Interval at which to resync contacts (in seconds).
     periodic_sync: 0
+    # Should leaving the room on Matrix make the user leave on Telegram?
```
sumnerevans, line 197:
```suggestion
    # Should leaving the room on Matrix make the user leave on Signal?
```

tulir (approved):

PR #246, opened 2022-04-17
Invite member

maltee1:
Tested with Signal only user and matrix user logged into the bridge

### DIFF
--- a/mautrix_signal/matrix.py
+++ b/mautrix_signal/matrix.py
@@ -17,7 +17,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
-from mautrix.bridge import BaseMatrixHandler
+from mautrix.bridge import BaseMatrixHandler, RejectMatrixIntvite
 from mautrix.types import (
     Event,
     EventID,
@@ -58,11 +58,16 @@ class MatrixHandler(BaseMatrixHandler):
         self, room_id: RoomID, user_id: UserID, inviter: u.User, event_id: EventID
     ) -> None:
         user = await u.User.get_by_mxid(user_id, create=False)
-        if not user:
+        if not user or not await user.is_logged_in():
             return
         portal = await po.Portal.get_by_mxid(room_id)
-        if portal:
-            await portal.handle_matrix_invite(inviter, user)
+        if portal and not portal.is_direct():
+            try:
+                await portal.handle_matrix_invite(inviter, user)
+            except RejectMatrixIntvite as e:
+                await portal.main_intent.send_notice(
+                    portal.mxid, f"Failed to invite {user.mxid} on Signal: {e}"
+                )
 
     async def send_welcome_message(self, room_id: RoomID, inviter: u.User) -> None:
         await super().send_welcome_message(room_id, inviter)
--- a/mautrix_signal/matrix.py
+++ b/mautrix_signal/matrix.py
@@ -61,7 +61,7 @@ class MatrixHandlerin(BaseMatrixHandler):
         if not user or not await user.is_logged_in():
             return
         portal = await po.Portal.get_by_mxid(room_id)
-        if portal and not portal.is_direct():
+        if portal and not portal.is_direct:
             try:
                 await portal.handle_matrix_invite(inviter, user)
             except RejectMatrixInvite as e:
--- a/mautrix_signal/matrix.py
+++ b/mautrix_signal/matrix.py
@@ -60,7 +60,6 @@ class MatrixHandler(BaseMatrixHandler):
         user = await u.User.get_by_mxid(user_id, create=False)
         if not user:
             return
-        await user.ensure_started()
         portal = await po.Portal.get_by_mxid(room_id)
         if portal:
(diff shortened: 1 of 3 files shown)


mautrix_signal/portal.py
```diff
@@ -722,6 +722,17 @@ async def handle_matrix_leave(self, user: u.User) -> None:
             self.log.debug(f"{user.mxid} left portal to {self.chat_id}")
             # TODO cleanup if empty
 
+    async def handle_matrix_invite(self, invited_by: u.User, user: u.User) -> None:
```
tulir, line 725:
```suggestion
    async def handle_matrix_invite(self, invited_by: u.User, user: u.User | p.Puppet) -> None:
```

mautrix_signal/matrix.py
```diff
@@ -54,6 +54,16 @@ def __init__(self, bridge: "SignalBridge") -> None:
...
 
+    async def handle_invite(
+        self, room_id: RoomID, user_id: UserID, inviter: u.User, event_id: EventID
+    ) -> None:
+        user = await u.User.get_by_mxid(user_id, create=False)
+        if not user:
+            return
+        portal = await po.Portal.get_by_mxid(room_id)
+        if portal:
+            await portal.handle_matrix_invite(inviter, user)
```
tulir, line 65: This should probably catch `RejectMatrixInvite` and send an error message (or at least log and ignore rather than letting it propagate). Something like this:

```suggestion
        if portal and not portal.is_direct:
            try:
                await portal.handle_matrix_invite(inviter, user)
            except RejectMatrixInvite as e:
                await portal.main_intent.send_notice(portal.mxid, "Failed to invite {user.mxid} on Signal: {e}")
```

(needs an import for `RejectMatrixInvite`)

mautrix_signal/portal.py
```diff
@@ -724,14 +724,18 @@ async def handle_matrix_leave(self, user: u.User) -> None:
...
         else:
             try:
                 await self.signal.update_group(
                     invited_by.username, self.chat_id, add_members=[user.address]
                 )
             except RPCError as e:
-                raise RejectMatrixInvite(e.message) from e
+                await self.main_intent.send_notice(
+                    self.mxid, f"Failed to invite {user.mxid} on Signal: {e}"
+                )
```
tulir, line 738: `RejectMatrixInvite` is the correct error for ghost invites, it's just the non-bridge-ghost `handle_invite` which wasn't catching them

mautrix_signal/matrix.py
```diff
@@ -42,7 +42,7 @@
     from .__main__ import SignalBridge
 
 
-class MatrixHandler(BaseMatrixHandler):
+class MatrixHandlerin(BaseMatrixHandler):
```
tulir, line 45: :thinking:

mautrix_signal/matrix.py
```diff
@@ -67,11 +82,9 @@ async def handle_leave(self, room_id: RoomID, user_id: UserID, event_id: EventID
         portal = await po.Portal.get_by_mxid(room_id)
         if not portal:
             return
-
```
sumnerevans, line 70: please revert these unnecessary whitespace changes

mautrix_signal/matrix.py
```diff
@@ -54,6 +54,21 @@ def __init__(self, bridge: "SignalBridge") -> None:
 
         super().__init__(bridge=bridge)
 
+    async def handle_invite(
```
sumnerevans, line 57: I don't think that you need this since `handle_puppet_invite` in https://github.com/mautrix/python/blob/master/mautrix/bridge/matrix.py#L306 should handle invites for you and call the portal function automatically, but I'm not 100% sure if that's how it works.

mautrix_signal/portal.py
```diff
@@ -725,6 +725,17 @@ async def handle_matrix_leave(self, user: u.User) -> None:
...
+    async def handle_matrix_invite(self, invited_by: u.User, user: u.User | p.Puppet) -> None:
+        if self.is_direct:
+            raise RejectMatrixInvite("You can't invite additional users to private chats.")
+        else:
+            try:
+                await self.signal.update_group(
+                    invited_by.username, self.chat_id, add_members=[user.address]
+                )
+            except RPCError as e:
+                raise RejectMatrixInvite(e.message) from e
```
sumnerevans, line 737:
```suggestion

        try:
            await self.signal.update_group(
                invited_by.username, self.chat_id, add_members=[user.address]
            )
        except RPCError as e:
            raise RejectMatrixInvite(e.message) from e
```

sumnerevans (dismissed):

mautrix_signal/matrix.py
tulir: This is for handling invites to users (rather than ghosts), so it is necessary currently

tulir (approved):

mautrix_signal/matrix.py
maltee1: done

PR #250, opened 2022-04-19
add basic create command

maltee1:
There are still a few problems. I'm not sure if they are caused by or just triggered by my changes
1. If I invite a ghost before running `create`, I will get a message that I already have a private chat with the user and the ghost leaves the room. I'm not sure how this is handled on the telegram bridge. IMHO, automatic dm creation should be limited to matrix rooms marked as direct. Inviting ghosts to rooms marked as rooms may trigger group creation in the future.
2. if I invite a ghost when there are already users in the room, the ghost joins directly. Once I run `create`, that ghost leaves the room.
3. I haven't been able to figure out where to obtain the matrix room's avatar url, so for now the room is created without an avatar.

member_role is currently set to DEFAULT. I suppose there isn't really a need to expose ADMINISTRATOR as an option. It's possibly somewhat related to a trusted private chat, but according to my understanding, the latter makes everyone who may join that room in the future automatically admin and the former only does that for the initial members. It's probably best to just map the power levels from the matrix side to signal 

edit:
2 has changed with the fix for the invite implementation. The ghost still joins immediately but is no longer kicked. It also isn't added to the group upon group creation.


maltee1:
OK, now it works
1. shouldn't have anything to do with the implementation of the command
2. works
3. still doesn't work, but can be added later imho.

mausignald/signald.py
```diff
@@ -389,6 +389,29 @@ async def get_group(
...
+        create_params = {
+            key: value
+            for key, value in {
+                "avatar": avatar_path,
+                "member_role": "ADMINISTRATOR" if member_role_administrator else "DEFAULT",
+                "title": title,
+                "members": [addr.serialize() for addr in members],
+            }.items()
+            if value is not None
+        }
```
sumnerevans, line 409: I think it's better to do the filter on a separate line.

```suggestion
        create_params = {
            "avatar": avatar_path,
            "member_role": "ADMINISTRATOR" if member_role_administrator else "DEFAULT",
            "title": title,
            "members": [addr.serialize() for addr in members],
        }
        create_params = {k: v for k, v in create_params.items() if v}
```

mautrix_signal/portal.py
```diff
@@ -1237,6 +1237,40 @@ async def handle_signal_delete(self, sender: p.Puppet, message_ts: int) -> None:
         except MForbidden:
             await self.main_intent.redact(message.mx_room, message.mxid)
 
+    # endregion
+    # region Matrix -> Signal metadata
+
+    async def create_signal_group(self, source: u.User) -> None:
+
```
sumnerevans, line 1244:
```suggestion
```

mautrix_signal/portal.py
```diff
@@ -1237,6 +1237,40 @@ async def handle_signal_delete(self, sender: p.Puppet, message_ts: int) -> None:
...
+        user_mxids = await self.az.intent.get_room_members(
+            self.mxid, (Membership.JOIN, Membership.INVITE)
+        )
+        invitee_addresses = list()
```
sumnerevans, line 1248:
```suggestion
        invitee_addresses = []
```

mautrix_signal/commands/util.py
```diff
@@ -0,0 +1,58 @@
+# mautrix-signal - A Matrix-Signal puppeting bridge
```
sumnerevans, line 1: neither of the functions in this file are used outside of `commands/signal.py`. There doesn't seem to be any reason to have them in a separate util module.

sumnerevans (dismissed):

mausignald/signald.py
maltee1: I have no opinion on that, I just copied it from somewhere else. I suppose it was done like that for a reason?

maltee1: the "update_group" request follows the same pattern for its parameters. I suppose we should either change both or leave this one the way it is.

sumnerevans: I think the two-statement way is clearer what's going on. Let's change in both places.

mautrix_signal/commands/signal.py
```diff
@@ -287,3 +288,77 @@ async def raw(evt: CommandEvent) -> None:
...
+            pass
+    return title, about, levels, encrypted, avatar_url
+
+
+async def warn_missing_power(levels: PowerLevelStateEventContent, evt: CommandEvent) -> None:
+    if levels.get_user_level(evt.az.bot_mxid) < levels.redact:
+        await evt.reply(
+            "Warning: The bot does not have privileges to redact messages on Matrix. "
+            "Message deletions from Signal will not be bridged unless you give "
+            f"redaction permissions to [{evt.az.bot_mxid}](https://matrix.to/#/{evt.az.bot_mxid})"
```
tulir, line 363: This error isn't applicable here since Signal only allows the sender to delete messages (and generally doesn't try to hide who deleted a message). Checking for invite and/or state default levels with a better message might still make sense though

mautrix_signal/commands/signal.py
```diff
@@ -287,3 +288,77 @@ async def raw(evt: CommandEvent) -> None:
...
+)
+async def create(evt: CommandEvent) -> EventID:
+    if await po.Portal.get_by_mxid(evt.room_id):
+        return await evt.reply("This is already a portal room.")
+
+    title, about, levels, encrypted, avatar_url = await get_initial_state(
+        evt.az.intent, evt.room_id
+    )
+    #    if not title:
+    #        return await evt.reply("Please set a title before creating a Telegram chat.")
```
tulir, line 307:
```suggestion
```

Should be uncommented or removed (is signal fine with unnamed group chats? :thinking:)

mausignald/signald.py
maltee1: I figured out that these two statements are not equivalent. The version I originally used seems to create an empty members list if there are no members, the one you suggested entirely omits the members command. The former is permitted, the latter is not. Same for an empty title.

tulir: The difference is `if v` vs `if value is not None`

maltee1: I see. I will adapt it accordingly. Being able to create empty groups and groups without a title should be possible if signal allows it.

mautrix_signal/commands/util.py
maltee1: done

mautrix_signal/commands/signal.py
```diff
@@ -287,3 +288,81 @@ async def raw(evt: CommandEvent) -> None:
...
+        receiver="",
+        avatar_url=avatar_url,
+    )
+
+    if levels.get_user_level(evt.az.bot_mxid) < 51:
+        await evt.reply(
+            "Warning: The bot does not have privileges to demote moderators on Matrix. "
+            "Demotions from ADMINISTRATOR to DEFAULT will not be bridged unless you give "
+            f"redaction permissions to [{evt.az.bot_mxid}](https://matrix.to/#/{evt.az.bot_mxid})"
+        )
```
tulir, line 322: Might be a good idea to check that the bot has a high enough level to change power levels in general, it might require PL 100

(also the message says redaction permissions)

mautrix_signal/commands/signal.py
```diff
@@ -287,3 +288,81 @@ async def raw(evt: CommandEvent) -> None:
...
+            "Warning: Permissions for changing name, topic and avatar cannot be set separately on Signal."
+            "name, room and avatar changes will not be bridged properly, unless those permissions are"
+            "set to the same level or lower than state_default"
+        )
+
+    try:
+        await portal.create_signal_group(evt.sender, levels)
+    except ValueError as e:
+        await portal.delete()
+        return await evt.reply(e.args[0])
```
tulir, line 338: I don't think this catch does anything useful

mautrix_signal/portal.py
```diff
@@ -1237,6 +1237,68 @@ async def handle_signal_delete(self, sender: p.Puppet, message_ts: int) -> None:
...
+                    AccessControlMode.MEMBER
+                    if levels.state_default == 0
+                    else AccessControlMode.ADMINISTRATOR
+                ),
+                members=None,
+                link=None,
+            ),
+        )
+        await self._postinit()
+        await self.insert()
```
tulir, line 1299: These should probably be immediately after the `create_group` call (it might also need some locking to make sure another portal isn't created)

mautrix_signal/commands/signal.py
```diff
@@ -287,3 +288,81 @@ async def raw(evt: CommandEvent) -> None:
...
+
+
+@command_handler(
+    needs_auth=True,
+    management_only=False,
+    help_section=SECTION_SIGNAL,
+    help_text=("Create a Signal group for the current Matrix room. "),
+)
+async def create(evt: CommandEvent) -> EventID:
+    if await po.Portal.get_by_mxid(evt.room_id):
```
tulir, line 300:
```suggestion
    if evt.portal:
```

mautrix_signal/portal.py
```diff
@@ -1237,6 +1237,68 @@ async def handle_signal_delete(self, sender: p.Puppet, message_ts: int) -> None:
...
 
+    # endregion
+    # region Matrix -> Signal metadata
+
+    async def create_signal_group(
+        self, source: u.User, levels: PowerLevelStateEventContent
+    ) -> None:
+        user_mxids = await self.az.intent.get_room_members(
+            self.mxid, (Membership.JOIN, Membership.INVITE)
+        )
```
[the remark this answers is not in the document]
maltee1: I copied this part from the telegram bridge. Maybe there's a reason it was done like that?

tulir (approved):

PR #257, opened 2022-05-09
add support for kick/ban/unban

maltee1:
Seems to work just fine.

We might have to think about power levels for kicking/banning/unbanning, because on signal, an `ADMINISTRATOR` can perform those actions on another `ADMINISTRATOR`, whereas matrix does not allow this if power levels are the same. I was thinking we could give a power level of 51 to matrix users, so they will be able to kick puppets, while Signal users can kick matrix users through the bridge bot.

### DIFF
--- a/mautrix_signal/matrix.py
+++ b/mautrix_signal/matrix.py
@@ -119,7 +119,7 @@ class MatrixHandler(BaseMatrixHandler):
                 await portal.unbridge()
             return
 
-        sender = await u.User.get_by_mxid(sender, create=False)
+        sender = await u.User.get_by_mxid(sender)
         if not await sender.is_logged_in() and portal.has_relay:
             sender = portal.relay_user_id
         if not sender:
--- a/mautrix_signal/matrix.py
+++ b/mautrix_signal/matrix.py
@@ -138,13 +138,11 @@ class MatrixHandler(BaseMatrixHandler):
     async def handle_kick(
         self, room_id: RoomID, user_id: UserID, kicked_by: UserID, reason: str, event_id: EventID
     ) -> None:
-
         await self.handle_kick_ban("kicked", room_id, user_id, kicked_by, reason, event_id)
 
     async def handle_unban(
         self, room_id: RoomID, user_id: UserID, unbanned_by: UserID, reason: str, event_id: EventID
     ) -> None:
-        # TODO handle unbans properly instead of handling it as a kick
         await self.handle_kick_ban("unbanned", room_id, user_id, unbanned_by, reason, event_id)
 
     async def handle_ban(
--- a/mautrix_signal/matrix.py
+++ b/mautrix_signal/matrix.py
@@ -120,11 +120,11 @@ class MatrixHandler(BaseMatrixHandler):
             return
 
         sender = await u.User.get_by_mxid(sender, create=False)
+        if not await sender.is_logged_in() and portal.has_relay:
+            sender = portal.relay_user_id
         if not sender:
             return
 
-        if not await sender.is_logged_in() and portal.has_relay:
-            sender = portal.relay_user_id
         user = await pu.Puppet.get_by_mxid(user_id)
         if not user:
             user = await u.User.get_by_mxid(user_id, create=False)
--- a/mautrix_signal/matrix.py
+++ b/mautrix_signal/matrix.py
@@ -120,8 +120,7 @@ class MatrixHandler(BaseMatrixHandler):
             return
 
         sender = await u.User.get_by_mxid(sender)
-        if not await sender.is_logged_in() and portal.has_relay:
-            sender = portal.relay_user_id
+        sender, is_relay = await self.get_relay_sender(sender, "kick/ban")
         if not sender:
             return
 
--- a/mautrix_signal/matrix.py
+++ b/mautrix_signal/matrix.py
@@ -100,6 +100,58 @@ class MatrixHandler(BaseMatrixHandler):
 
         await portal.handle_matrix_join(user)
 
+    async def handle_kick_ban(
+        self,
+        action: str,
+        room_id: RoomID,
+        user_id: UserID,
+        sender: UserID,
+        reason: str,
+        event_id: EventID,
+    ) -> None:
+        self.log.debug(f"{user_id} was {action} from {room_id} by {sender} for {reason}")
+        portal = await po.Portal.get_by_mxid(room_id)
+        if not portal:
+            return
+
+        if user_id == self.az.bot_mxid:
+            if portal.is_direct:
+                await portal.cleanup_and_delete()
+            return
+
+        sender = await u.User.get_by_mxid(sender, create=False)
+        if not sender:
+            return
+
+        user = await pu.Puppet.get_by_mxid(user_id)
+        if not user:
+            user = await u.User.get_by_mxid(user_id, create=False)
+            if not user:
+                return
+        if action == "banned":
+            await portal.ban_matrix(user, sender)
+        elif action == "kicked":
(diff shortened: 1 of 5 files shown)


mautrix_signal/portal.py
```diff
@@ -725,6 +725,43 @@ async def handle_matrix_leave(self, user: u.User) -> None:
...
+        except Exception as e:
+            self.log.exception(f"Failed to ban Signal user: {e}")
+            await self.main_intent.unban_user(self.mxid, user.mxid)
+            info = await self.signal.get_group(source.username, self.chat_id)
+            if user.address in info.members:
+                await self.main_intent.invite_user(
+                    self.mxid,
+                    user.mxid,
+                    check_cache=True,
+                    reason=f"Failed to ban Signal user: {e}",
```
maltee1, line 752: I put the reason here because unbans don't come with reasons. If the user wasn't in the room at the time of the failed ban attempt, the error will not be printed to the room. I don't think it's important enough to have the bridge bot send a notification in that case.

tulir: Unbans can have reasons (might need to update mautrix-python) so it should just be put there

mautrix_signal/matrix.py
```diff
@@ -100,6 +100,56 @@ async def handle_join(self, room_id: RoomID, user_id: UserID, event_id: EventID)
...
+        if not portal:
+            return
+
+        if user_id == self.az.bot_mxid:
+            if portal.is_direct:
+                await portal.cleanup_and_delete()
+            return
+
+        sender = await u.User.get_by_mxid(sender, create=False)
+        if not sender:
```
tulir, line 123: This should check if the user is logged in (or alternatively it should support using the relaybot)

```suggestion
        if not sender or not await sender.is_logged_in():
```

mautrix_signal/matrix.py
```diff
@@ -100,6 +100,56 @@ async def handle_join(self, room_id: RoomID, user_id: UserID, event_id: EventID)
...
+            return
+
+        user = await pu.Puppet.get_by_mxid(user_id)
+        if not user:
+            user = await u.User.get_by_mxid(user_id, create=False)
+            if not user:
```
tulir, line 129: This one should always skip non-logged-in users even if relaybots are supported

```suggestion
            if not user or not await user.is_logged_in():
```

mautrix_signal/portal.py
```diff
@@ -725,6 +725,43 @@ async def handle_matrix_leave(self, user: u.User) -> None:
...
+        try:
+            await self.signal.update_group(
+                source.username, self.chat_id, remove_members=[user.address]
+            )
+        except Exception as e:
+            self.log.exception(f"Failed to kick Signal user: {e}")
+            await self.main_intent.invite_user(
+                self.mxid, user.mxid, check_cache=True, reason=f"Failed to kick Signal user: {e}"
+            )
+            await user.intent_for(self).ensure_joined(self.mxid)
```
tulir, line 738: Hmm, would update_group throw an error if the user is not present in the group? (i.e. will kicking a ghost who isn't in the signal chat cause the ghost to rejoin, even though they aren't in the chat?)

mautrix_signal/matrix.py
```diff
@@ -100,6 +100,56 @@ async def handle_join(self, room_id: RoomID, user_id: UserID, event_id: EventID)
...
+    ) -> None:
+        self.log.debug(f"{user_id} was {action} from {room_id} by {sender} for {reason}")
+        portal = await po.Portal.get_by_mxid(room_id)
+        if not portal:
+            return
+
+        if user_id == self.az.bot_mxid:
+            if portal.is_direct:
+                await portal.cleanup_and_delete()
+            return
```
tulir, line 120: This part looks a bit weird.

DMs should never go to the other parts of the code and I'm not sure if kicking the bridge bot should unconditionally delete the DM (it's only required for e2ee, and the recipient is the main_intent in DM portals). The bot is required for group chats though, so unbridging when the bridge bot is kicked from a group would make sense.

mautrix_signal/portal.py
maltee1: good point. Same goes for ban and unban. We should probably check the signal group in all three cases.

mautrix_signal/portal.py
maltee1: It doesn't look like it:
https://github.com/mautrix/python/blob/4ba953d7180bfc0b2d8b822c0701111796b9d6df/mautrix/client/store_updater.py#L113
https://matrix.org/docs/spec/client_server/r0.6.1.html#post-matrix-client-r0-rooms-roomid-unban

tulir: https://spec.matrix.org/v1.2/client-server-api/#post_matrixclientv3roomsroomidunban

maltee1: apparently I checked the wrong version. I changed it, but I currently can't see the reasons in my client. I'm not sure where the problem lies (most likely in my mautrix-python changes), but I don't think it's here.

maltee1: Correction, it works with the pull request for mautrix-python. Would I have to adjust the version for mautrix-python somewhere in the requirements?

maltee1: works

mautrix_signal/matrix.py
```diff
@@ -100,6 +100,58 @@ async def handle_join(self, room_id: RoomID, user_id: UserID, event_id: EventID)
...
+
+        if user_id == self.az.bot_mxid:
+            if portal.is_direct:
+                await portal.unbridge()
+            return
+
+        sender = await u.User.get_by_mxid(sender, create=False)
+        if not await sender.is_logged_in() and portal.has_relay:
+            sender = portal.relay_user_id
```
tulir, line 124: Same issue as https://github.com/mautrix/signal/pull/260#discussion_r885029219, but since none of the new methods here have relaybot handling in the portal, it might be fine to put it here

```suggestion
        sender, is_relay = await self.get_relay_sender(sender, "kick/ban")
```

mautrix_signal/matrix.py
```diff
@@ -100,6 +100,58 @@ async def handle_join(self, room_id: RoomID, user_id: UserID, event_id: EventID)
...
+        portal = await po.Portal.get_by_mxid(room_id)
+        if not portal:
+            return
+
+        if user_id == self.az.bot_mxid:
+            if portal.is_direct:
+                await portal.unbridge()
+            return
+
+        sender = await u.User.get_by_mxid(sender, create=False)
```
tulir, line 122: This probably shouldn't have `create=False` since creating the `User` instance is required even when using the relaybot

```suggestion
        sender = await u.User.get_by_mxid(sender)
```

mautrix_signal/portal.py
```diff
@@ -725,6 +725,52 @@ async def handle_matrix_leave(self, user: u.User) -> None:
...
+                )
+                await user.intent_for(self).ensure_joined(self.mxid)
+
+    async def unban_matrix(self, user: u.User | p.Puppet, source: u.User) -> None:
+        try:
+            await self.signal.unban_user(source.username, self.chat_id, users=[user.address])
+        except Exception as e:
+            self.log.exception(f"Failed to unban Signal user: {e}")
+            info = await self.signal.get_group(source.username, self.chat_id)
+            if info.banned_members and user.address in info.banned_members:
```
tulir, line 769: These `in banned_members` checks are still wrong

maltee1: OK, I did some more testing ((un)banning with insufficient rights, (un)banning when the user already is (un)banned). Everything works as expected.

tulir (approved):

PR #260, opened 2022-05-22
add support for matrix -> signal power level changes

maltee1:
make Signal role ADMINISTRATOR if power level >= 50
make Signal role DEFAULT if power level <50
bridge `invite` and `state_default` upon explicit changes

works for group creation, invites or explicit power level changes

### DIFF
--- a/mautrix_signal/portal.py
+++ b/mautrix_signal/portal.py
@@ -36,6 +36,7 @@ from mausignald.types import (
     Group,
     GroupAccessControl,
     GroupID,
+    GroupMember,
     GroupMemberRole,
     GroupV2,
     GroupV2ID,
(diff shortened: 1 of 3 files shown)


maltee1:
I'm catching exceptions separately for each user and the permissions. I'm thinking whether that's useful, because any exception that will occur for one of the requests will likely occur for the other requests as well. So putting a try/catch block around the entire function might make sense.

mautrix_signal/portal.py
```diff
@@ -800,6 +815,95 @@ async def handle_matrix_avatar(self, user: u.User, url: ContentURI) -> None:
...
+                            if levels.state_default == 0
+                            else AccessControlMode.ADMINISTRATOR
+                        ),
+                        members=None,
+                        link=None,
+                    ),
+                )
+                self.revision = update_meta.revision
+            except Exception as e:
+                self.log.exception(f"Failed to update Signal GroupAccessControl: {e}")
```
tulir, line 902:
```suggestion
                self.log.exception(f"Failed to update Signal metadata change permission: {e}")
```

mautrix_signal/portal.py
```diff
@@ -800,6 +815,95 @@ async def handle_matrix_avatar(self, user: u.User, url: ContentURI) -> None:
...
+                            if levels.invite == 0
+                            else AccessControlMode.ADMINISTRATOR
+                        ),
+                        attributes=None,
+                        link=None,
+                    ),
+                )
+                self.revision = update_meta.revision
+            except Exception as e:
+                self.log.exception(f"Failed to update Signal GroupAccessControl: {e}")
```
tulir, line 881:
```suggestion
                self.log.exception(f"Failed to update Signal member add permission: {e}")
```

tulir (approved):

PR #263, opened 2022-05-26
tolerate ranges of power levels for logged-in users to avoid conflicts

### LINKED ISSUE #258: Tolerate ranges of power levels for logged-in matrix users
Enforcing pl 0 or 50 makes sense for signal only groups, but it's problematic once groups are mixed. Users might want more fine-grained control for platforms other than Signal, without the signal bridge interfering with that. Also, if only a single user cannot be demoted (because it has the same or higher power level as the bridge bot) the entire request will fail and no power levels will be updated. Having pl 100 in a cross-platform group you created is reasonable, but will block any Signal -> Matrix power level changes, because promoting the bridge bot to anything higher is not possible (and not necessarily desirable).

I'm not sure about a sane way of handling this, that's why I'm opening an issue. I'm thinking it would make sense to
1. enforce power levels as before for Signal puppets
2. For matrix users, don't change matrix power levels within a range, e.g. don't demote Signal `DEFAULT` if `0<matrix pl< 50`, don't demote Signal `ADMINISTRATOR`. Do demote `DEFAULT` if `matrix pl > 50` and do promote `ADMINSTRATOR` if `matrix pl < 50`.
3. Do not try to change power levels of users with the same or higher power level as the bridge bot

maltee1:
Fixes #258

for logged-in users:
for DEFAULT, the power level is between 0 and 49
for ADMINISTRATOR, the power level is >50
The bot will only change the power level if it is outside those ranges.

for signal puppets, nothing changes.

If the current power level of the user is higher than that of the bridge bot, don't attempt to change it. While the power level may have to be changed, it's pointless to try.
Likewise, if the bot power level is <50, it won't be able to do anything useful and it's pointless to try. We already print a warning for that.

This PR avoids conflicts with other bridges trying to set different power levels (pl 75 is default for the group creator in telegram, in whatsapp it's 95). It will also allow setting power levels even if they can't be set for some users. Previously, a single, impossible power level change would make the entire command fail, causing no power level changes to be made.

### DIFF
--- a/mautrix_signal/portal.py
+++ b/mautrix_signal/portal.py
@@ -1679,6 +1679,7 @@ class Portal(DBPortal, BasePortal):
         is_initial: bool = False,
     ) -> PowerLevelStateEventContent:
         levels = levels or PowerLevelStateEventContent()
+        bot_pl = levels.get_user_level(self.az.bot_mxid)
         levels.events_default = 0
         if self.is_direct:
             levels.ban = 99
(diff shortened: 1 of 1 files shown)


mautrix_signal/portal.py
```diff
@@ -1691,8 +1692,20 @@ async def _get_power_levels(
                 ac = info.access_control
                 for detail in info.member_detail + info.pending_member_detail:
                     puppet = await p.Puppet.get_by_address(Address(uuid=detail.uuid))
-                    level = 50 if detail.role == GroupMemberRole.ADMINISTRATOR else 0
-                    levels.users[puppet.intent_for(self).mxid] = level
+                    puppet_mxid = puppet.intent_for(self).mxid
+                    current_level = levels.get_user_level(puppet_mxid)
+                    if bot_pl > current_level and bot_pl >= 50:
+                        level = current_level
+                        if await u.User.get_by_address(Address(uuid=detail.uuid)):
```
tulir, line 1699: Seems like it would be easier to just check if the mxid is a double puppet:

```suggestion
                        if puppet.is_real_user:
```

tulir (approved):

PR #266, opened 2022-06-01
Remove duplicate blurhash key

Detegr:
According to the blurhash [proposal](https://github.com/matrix-org/matrix-spec-proposals/blob/anoa/blurhash/proposals/2448-blurhash-for-media.md):
> Implementations wishing to add
> this before this MSC is merged can do so with the following: The
> blurhash key in any events, request or response bodies should be
> replaced with xyz.amorgan.blurhash.

The current implementation of mautrix-signal sets both `blurhash` and
`xyz.amorgan.blurhash`, which breaks for example the deserialization of
such events in [Ruma](https://github.com/ruma/ruma). Ruma handles `xyz.amorgan.blurhash` as
`blurhash`, and when there's also a `blurhash` key present, it fails to
deserialize the event with "duplicate field" error.

This commit fixes the issue by just setting the `xyz.amorgan.blurhash`
key, as per the instructions in the proposal.

### DIFF
--- a/mautrix_signal/portal.py
+++ b/mautrix_signal/portal.py
@@ -916,7 +916,6 @@ class Portal(DBPortal, BasePortal):
 
             if content:
                 if message.sticker.attachment.blurhash:
-                    content.info["blurhash"] = message.sticker.attachment.blurhash
                     content.info["xyz.amorgan.blurhash"] = message.sticker.attachment.blurhash
                 await self._add_sticker_meta(message.sticker, content)
                 if reply_to and not message.body:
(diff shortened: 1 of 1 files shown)


tulir:
I usually prefer sending stable prefixes right away to avoid unnecessarily breaking old events when the MSC is merged.

That specific MSC is also nearly ready, the only remaining concern is server implementation of profile endpoints, nothing related to the event content.

Detegr:
So are you suggesting to remove the xyz.amorgan.blurhash and leave the blurhash key instead?

Detegr:
But the issue is that it is against the suggested way of implementing the feature, and for example the newest version of Ruma will fail to parse messages that have the blurhash twice.
The way I see it Ruma works according to the proposal and mautrix-signal violates it. I get the appeal of using both keys but clearly it is not working with every client.

sumnerevans:
Ruma should be more resilient to random keys being present in `content`.

PR #270, opened 2022-06-11
catch ProfileUnavailableError in _update_participants

maltee1:
currently, a ProfileUnavailableError will cause _update_participants to fail and groups won't update properly. This PR catches the Error so the remainder of _update_participants can execute.


tulir (approved):

TheDodger:
Bump. 😅

PR #272, opened 2022-06-16
bridge signal -> matrix power level changes with correct sender if possible

maltee1:
If the puppet has sufficient rights, bridge power levels changes with the puppet. Otherwise use the bridge bot. This will typically mean that promotions to admin will be bridged as coming from the correct sender while demotions will be bridged by the bot.

This also includes a fix to stop the bot from trying to set its power levels to 100, because there seems to be no good reason to do so. If the bot has pl <100, the request will fail, if it has pl >100, it will demote itself, but there is no good reason to do that. Instead we might try to look for a double-puppeted user with pl 100 to promote the bridge bot or just print a warning.
Initial power level for bot-created rooms should remain 9001.
I haven't found an elegant solution for not sending an event if power levels haven't changed. Nheko prints a "xyz has changed the room permissions" in that case which I guess is a nheko bug (because nothing has changed), but we might also not send an event if nothing has changed. This was a problem before already, but it's exacerbated by this PR.

### DIFF
--- a/mautrix_signal/portal.py
+++ b/mautrix_signal/portal.py
@@ -1500,7 +1500,7 @@ class Portal(DBPortal, BasePortal):
         changed = await self._update_avatar(info, sender) or changed
         await self._update_participants(source, info)
         try:
-            await self._update_power_levels(info)
+            await self._update_power_levels(info, sender)
         except Exception:
             self.log.warning("Error updating power levels", exc_info=True)
         if changed:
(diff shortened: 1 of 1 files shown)


mautrix_signal/portal.py
```diff
@@ -1677,11 +1677,16 @@ async def _update_participants(self, source: u.User, info: ChatInfo) -> None:
...
+    async def _update_power_levels(self, info: ChatInfo, sender: p.Puppet | None = None) -> None:
         if not self.mxid:
             return
 
         power_levels = await self.main_intent.get_power_levels(self.mxid)
+        if sender:
+            power_levels = await self._get_power_levels(
+                power_levels, info=info, is_initial=False, sender=sender
+            )
+            await sender.intent_for(self).set_power_levels(self.mxid, power_levels)
```
sumnerevans, line 1689: I think you want an early return here or an `else`, right?

maltee1: If there is a sender, `set_power_levels` must be run twice, once setting all power levels the puppet is allowed to send and again to set those the puppet isn't allowed to send by the bridge bot. Demotions from ADMINISTRATOR to DEFAULT for example can not be performed by the sender, since the sender and the demotee have the same power level (this is allowed on signal). In this case `if sender` will be executed but produce no change, the change must then be performed by the bridge bot.
What I mentioned in the PR description is that this runs set_power_levels unnecessarily quite a lot (usually either the command by the puppet or the command by the bridge bot will produce no change), so we might want to detect whether anything has changed before sending the request. I'm not really sure about a good way to do this, returning a `bool: changed` with `_get_power_levels` seems to be quite complicated, maybe we can pass a deep copy to `_get_power_levels` and compare before and after?

maltee1:
I followed your advice to serialize + deserialize. It works, but isn't it quite inefficient? I also tried `copy.deepcopy()`, but that gave me a bunch of errors that I didn't understand.

PR #273, opened 2022-06-17
bridge kick/invite/leave from Signal to Matrix with correct sender

maltee1:
This PR bridges kick/invite with correct metadata
- changes detected by sync/restart are performed by the bridge bot. Otherwise:
- invites are always performed by the correct puppet
- kicks are preferably performed by the correct puppet. If the user/puppet that is being kicked is an `ADMINISTRATOR`, then they can be kicked on Signal, but kicking users with equal power level is not permitted on matrix. In that case the kick is performed by bridge bot, with the sender given in the reason.
- The reason for kicking is now formulated to address everyone, not just the person being kicked (previously: "xyz kicked you", "you are not a member..."
- Only leaves are bridged as leaves.

If `levels.invite`>50, then signal `ADMINISTRATOR`s are still able to invite users, the invite is then bridged by the bot, but no reason is provided, so the inviter is unknown on the matrix side. Currently, `levels.invite` is forced to 50 or 0, so this shouldn't currently be an issue, but we should consider respecting the setting being changed in matrix for plumbed rooms.

### DIFF
--- a/mautrix_signal/portal.py
+++ b/mautrix_signal/portal.py
@@ -1153,7 +1153,7 @@ class Portal(DBPortal, BasePortal):
 
     async def handle_signal_kicked(self, user: u.User, sender: p.Puppet) -> None:
         self.log.debug(f"{user.mxid} was kicked by {sender.number} from {self.mxid}")
-        await self.main_intent.kick_user(self.mxid, user.mxid, f"{sender.name} kicked you")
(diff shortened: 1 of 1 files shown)


tulir (approved):
The except should catch specific errors and there are some merge conflicts, otherwise looks fine

PR #293, opened 2022-08-23
add relaybot config option and auto-set relay for relaybot

maltee1:
This PR allows users to set a dedicated relaybot user. When the relaybot is invited, it will automatically be set as the relay. This way, users no longer need to log in with the relaybot account to run set-relay manually. The relaybot can be shared between users more easily.

### DIFF
--- a/mautrix_signal/portal.py
+++ b/mautrix_signal/portal.py
@@ -863,6 +863,19 @@ class Portal(DBPortal, BasePortal):
             )
         except RPCError as e:
             raise RejectMatrixInvite(str(e)) from e
+        if user.mxid == self.config["bridge.relay.relaybot"]:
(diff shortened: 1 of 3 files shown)


mautrix_signal/portal.py
```diff
@@ -863,6 +863,19 @@ async def handle_matrix_invite(self, invited_by: u.User, user: u.User | p.Puppet
             )
         except RPCError as e:
             raise RejectMatrixInvite(str(e)) from e
+        if user.mxid == self.config["bridge.relay.relaybot"]:
```
tulir, line 866:
```suggestion
        if user.mxid == self.config["bridge.relay.relaybot"] != "@relaybot:example.com":
```

tulir (approved):

PR #295, opened 2022-08-27
set relaybot as relay on create

maltee1:
- move relaybot invite handling to separate function
- also set relaybot as relay when creating a new portal. I don't find the check for the bot in the for loop particularly pretty, but it seemed to be the most efficient option.
- also fixes a bug that breaks group creation because join_rule isn't passed by get_initial state.

### DIFF
--- a/mautrix_signal/portal.py
+++ b/mautrix_signal/portal.py
@@ -863,18 +863,7 @@ class Portal(DBPortal, BasePortal):
         except RPCError as e:
             raise RejectMatrixInvite(str(e)) from e
         if user.mxid == self.config["bridge.relay.relaybot"] != "@relaybot:example.com":
-            if not self.config["bridge.relay.enabled"]:
-                await self.main_intent.send_notice(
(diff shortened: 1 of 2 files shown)


mautrix_signal/portal.py
```diff
@@ -1756,10 +1758,14 @@ async def create_signal_group(
             self.mxid, (Membership.JOIN, Membership.INVITE)
         )
         invitee_addresses = []
+        relaybot_mxid = self.config["bridge.relay.relaybot"]
+        relaybot = None
         for mxid in user_mxids:
             mx_user = await u.User.get_by_mxid(mxid, create=False)
             if mx_user and mx_user.address and mx_user.username != source.username:
                 invitee_addresses.append(mx_user.address)
+                if mxid == relaybot_mxid and relaybot_mxid != "@relaybot:example.com":
```
tulir, line 1767:
```suggestion
                if mxid == relaybot_mxid != "@relaybot:example.com":
```

tulir (approved):

PR #320, opened 2022-11-09
Add ProofRequiredError Exception and submit-challenge command

maltee1:
This is still entirely untested since I would need a ProofRequiredError to do so. I am not currently commited enough to try and trigger one for my Signal account, but anyone who has one, please try. Or if you find obvious mistakes without testing, let me know.

I've noticed that ProofRequiredError rarely comes in the SendMessageResult, but usually as a ResponseError, so I created another ProofRequiredException in errors.py. I didn't know how to reconcile the two, since SendMessageResult probably wouldn't work with an Exception instead of a SerializableAttrs, but I also needed to inherit from ResponseError.

When portal.handle_matrix_message() encounters a ProofRequiredError, it should store a token in the respective user, so the user won't have to copy&paste that. Since tokens are temporary in nature and I don't know how to work with databases, it's not stored in the database. I suppose that's fine.

The user can then use the submit-challenge command with just a captcha token to complete the challenge. I currently don't handle PUSH_CHALLENGE. So far, I have only encountered RECAPTCHA, I would prefer to first handle that and then worry about the other (probably just two lines, but still).

### DIFF
--- a/mausignald/errors.py
+++ b/mausignald/errors.py
@@ -7,6 +7,8 @@ from __future__ import annotations
 
 from typing import Any
 
+from mautrix.util.format_duration import format_duration
+
 
 class RPCError(Exception):
     pass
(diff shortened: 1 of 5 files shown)


maltee1:
Tested & working. Please have a look.

tulir (approved):

exmatrikulator:
How do I get the captcha token?
Can you add a documentation for that?

Thanks

maltee1:
I can't, because the python version of the bridge is deprecated. I suppose the go version will at some point have a similar command, if captchas continue to be requested.

Here's the documentation:
https://signald.org/articles/captcha/

PR #326, opened 2022-11-27
Fix v11 migration for SQLite

crawford:
The double-colon syntax isn't supported by SQLite. When trying to migrate to v11, the following error is seen:

    sqlite3.OperationalError: unrecognized token: ":"

Instead, use the ANSI SQL syntax.


crawford:
Looks like 4839e2afbef6179f4e0b494298086ac802ff912c beat me to it.

PR #327, opened 2022-12-04
Add additional comment for clarification on domain setting

Fmstrat:
A common issue reported here and in support requests is the the failed `/registration`. It is currently covered in the Troubleshooting and FAQ here: https://docs.mau.fi/bridges/general/troubleshooting.html

For new users, the problem likely stems from the use of the term `domain` instead of `server_name`. Additionally, using `example.com` instead of something like `synapse.example.com` or `matrix.example.com` could be leading to more confusion since many people aren't setting up a matrix instance on their root domain.

This PR updates the example config to include the hostname with the domain, and adds an additional comment that should hopefully eliminate any future confusion.

### DIFF
--- a/mautrix_signal/example-config.yaml
+++ b/mautrix_signal/example-config.yaml
@@ -1,9 +1,10 @@
 # Homeserver details
 homeserver:
     # The address that this appservice can use to connect to the homeserver.
-    address: https://example.com
+    address: https://synapse.example.com
     # The domain of the homeserver (for MXIDs, etc).
-    domain: example.com
+    # This should be the exact same value as `server_name` in homeserver.yaml
+    domain: synapse.example.com
     # Whether or not to verify the SSL certificate of the homeserver.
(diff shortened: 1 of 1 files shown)


Fmstrat:
Hmm, the more I think about this, the tougher it is. Many new users will be following the guidance of using a subdomain instead of their main domain on first setup (https://matrix-org.github.io/synapse/latest/setup/installation.html), so if this change is to be made, it should probably be made throughout the config file.

However, the config file is built for a production release. Perhaps just the comment and nothing else?

tulir:
You generally shouldn't be using a subdomain as the server name, so the example domain should definitely not include one.

The example address field could maybe have a subdomain, but shouldn't be ` `synapse.` since there are other implementations too, and `matrix.` is more common anyway.

Fmstrat:
Agreed. Any objections to me revising to just update from a comment perspective?

PR #333, opened 2023-01-04
fix power level assignment for non-double-puppeted users, some cleanup

maltee1:
Getting a puppet for a double-puppeted user returns the user, if double puppeting is not enabled, it will only return their puppet. This caused non-double-puppeted users to not receive appropriate power level for their account.

This PR also cleans up a few loosely related things:
1. All remaining GroupV1 support is removed (hopefully)
2. ChatInfo is now limited to full group info, GroupV2ID is not allowed. I also replaced ChatInfo by GroupV2 in some places. Using ChatInfo for both GroupV2 and GroupV2ID made it necessary to always check for full group info and pass a source to obtain full group info, even if that ended up never being used. I suppose the same is true for Profile and Address, but they're just used a lot less, so it's less annoying

### DIFF
--- a/mautrix_signal/portal.py
+++ b/mautrix_signal/portal.py
@@ -37,7 +37,6 @@ from mausignald.types import (
     Address,
     AnnouncementsMode,
     Attachment,
-    Group,
     GroupAccessControl,
     GroupChange,
     GroupID,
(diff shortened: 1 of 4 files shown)


mautrix_signal/portal.py
```diff
@@ -2367,31 +2344,33 @@ async def _get_power_levels(
...
-                                level = 0
-                            elif (
-                                current_level < 50 and detail.role == GroupMemberRole.ADMINISTRATOR
-                            ):
-                                level = 50
-                        else:
-                            level = 50 if detail.role == GroupMemberRole.ADMINISTRATOR else 0
+            if not info:
+                self.log.debug(f"No info provided for updating power levels in {self.mxid}")
+                return
```
tulir, line 2349: Returning `None` from here is not allowed, but the case seems pointless anyway since info is always provided.

tulir (approved):

Issue #102, opened 2021-06-20
feature request: Spaces support


williamkray:
I currently use the Spaces beta, and it would be excellent to have support for Spaces added to this bridge to allow all signal-related messages lumped together similar to the existing Communities support.

MayeulC:
To detail more, I would like to:
* Add every "DM" and group chat to the space
* Add every contact to the space
  * even those the user hasn't spoken with, maybe as "suggested rooms". It could be a subspace, but I am not sure how supported  those are for now.
* Add the management DM with the signal bot to that space.

ShadowJonathan:
WhatsApp Mautrix has this now, so it should be relatively easy to replicate the feature

Tropaion:
> WhatsApp Mautrix has this now, so it should be relatively easy to replicate the feature
> 
> Edit: Wrong language, whoops

Wanted to recommend to replicate it from Discord Mautrix, but is also Go.
But I would really like the Space support, I'm using it with WhatsApp and Discord and is really nice.

Glotzon:
Yes, please, this feature is urgently missing. Maybe even the signal bridge should by default automatically generate a space and place everything underneath.

Tropaion:
I have just seen, heisenbridge is also written in python and already supports spaces: https://github.com/hifi/heisenbridge/blob/master/heisenbridge/space_room.py

Maybe this would make it easier to implement it. Sadly I don't know much about python.

Jaffex:
I adopted the [changes](https://github.com/mautrix/whatsapp/pull/413) made to the whatsapp bridge and added it to the signal bridge over at https://github.com/Jaffex/signal/tree/add/spaceGeneration

Currently however, a room would need to be recreated for it to be added on an already existing installation.

@tulir would you be interested in a pull request and would you have any advice on how to add already existing rooms to the space?

sumnerevans:
@Jaffex you can check out what I did in LinkedIn:

* https://github.com/beeper/linkedin/commit/e62020dcf219f3371f608ceb9d8e5d4acd47f505
* https://github.com/beeper/linkedin/commit/3d4afcd24b9fb9f87cd722e50a3a6d873dccc10d

Also, I think the preferred nomenclature is `personal_filtering_spaces` since that's what WhatsApp does: https://github.com/mautrix/whatsapp/blob/master/example-config.yaml#L105-L107

Issue #183, opened 2021-12-09
Discover and reuse existing DM rooms instead of creating new ones


MayeulC:
Some time ago, I had a few issues with the bridge, which led me to uninstall it.

Reinstalling it went fine, but it's re-creating every room. I think the bridge should be able to tell when a direct chat between a signal user and a matrix user already exists, and use that one.

The same issue also exists for group chats: a new one is created, the old one is left un-updated even though the bot user still appears as a member. It might be more complex to handle, especially if there are also Matrix users inside.

One possible solution would be to use signal uuids (possibly hashed, maybe peppered with server name) to build Matrix room ids deterministically.

tulir:
You can just use `clean-rooms` to clean up the old rooms

witchent:
Maybe they don't want to loose the message history, so I agree that it would be nice if it would reuse old rooms (or if one could manually make a room belong to a certain signal chat maybe)

MayeulC:
Yes, I would have liked to retain room history if possible, and space membership as well. To a lesser extent, Matrix users membership. It would have also been handy to contact users I was recently chatting with as these portals were already "open".

Another option to recognize rooms would be to put Signal metadata inside some custom event.

Thank you for mentioning `clean-rooms`, I had forgotten about it, and it wasn't clear to me that this is what it is for. I'll probably use it for now, but reusing the rooms would be nicer.

PR #334, opened 2023-01-04
Add personal space for signal chats

Jaffex:
As requested in #102:

- Adds a config option to enable personal filtering spaces
- When activated, creating or updating a room will add it to the user's personal space
- Upgrade compatibility by making the sync command add all group and private chats to the space
(Since sync seems to be automatically invoked upon startup, no additional update steps are needed after setting the config option)

Caveats
- The space child event is always sent upon matrix room update, even if the room already belongs to the space
- Rooms are not automatically removed from space upon unbridging (This might be desireable however, since the user will probably want to see archived unbridged chats still belonging to the Signal space - since they are automatically added as a mod to the space, they might remove the chat from it as they please)

### DIFF
--- a/mautrix_signal/config.py
+++ b/mautrix_signal/config.py
@@ -45,6 +45,7 @@ class Config(BaseBridgeConfig):
         copy("signal.remove_file_after_handling")
         copy("signal.registration_enabled")
         copy("signal.enable_disappearing_messages_in_groups")
+        copy("signal.space_per_user")
 
         copy("metrics.enabled")
         copy("metrics.listen_port")
--- a/mautrix_signal/config.py
+++ b/mautrix_signal/config.py
@@ -45,13 +45,13 @@ class Config(BaseBridgeConfig):
         copy("signal.remove_file_after_handling")
         copy("signal.registration_enabled")
         copy("signal.enable_disappearing_messages_in_groups")
-        copy("signal.space_per_user")
 
         copy("metrics.enabled")
         copy("metrics.listen_port")
 
         copy("bridge.username_template")
         copy("bridge.displayname_template")
+        copy("bridge.personal_filtering_spaces")
         if self["bridge.allow_contact_list_name_updates"]:
             base["bridge.contact_list_names"] = "allow"
         else:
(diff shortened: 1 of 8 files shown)


MayeulC:
I have a few questions as a user (sorry, this is not a proper review, as I am not familiar enough with the code):

1. Does `set-pl` work in the space room? And more generally, does the bot answer in the space room? That would allow stuff like change the room avatar, but some actions might be broken.
2. I personally keep unbridged rooms in a subspace or a different space. Could an option or space command be added to perform that (possibly in a follow-up PR). I'm sometimes struggling to identify which rooms are still bridged, due to the duplicates I have from #183.
3. Do all contacts get added to the space, even the ones that were never messaged? Similar to `autocreate_contact_portal`, but without inviting the user. That would make it easier to initiate a message from Matrix. If not included, I'd really love to see a follow-up PR :)

Jaffex:
Hi there,

1. I'm not sure there is any way to "talk" inside a space room, but since they are basically no different than regular rooms, I think `set-pl [mxid] powerlevel` should work.

 You are made a moderator by default however, so you'd be able to change the space avatar as well as add your own chats. 

The avatar, name, etc. only gets set at creation of the space so your changes should persist. (Only thing that would not persist: if you remove a room from the space, it will get added again upon the next sync command or restart of the bridge)

2. I guess everything is possible, but it's out of the scope of this PR :)

3. Rooms that get created (or updated) while the setting is active will always be put inside the space. So if `autocreate_contact_portal` is on, those rooms will be attached to the space.
The signal puppets themselves will not get added as members to the space.

MayeulC:
Thanks for the answers :)

> 1. I'm not sure there is any way to "talk" inside a space room

You sure can, even directly from element if you right click the space -> "Show room history (dev tools)". It might require a certain power level, of course :)

> 2.

:)

> 3. Rooms that get created (or updated) while the setting is active will always be put inside the space. So if autocreate_contact_portal is on, those rooms will be attached to the space.
The signal puppets themselves will not get added as members to the space.

That makes sense. Ideally there should be a setting to allow creating these portal rooms and add them to the space, but without inviting the user. That's probably material for another PR, and something that should be better adjusted on a per-user basis.

PR #392, opened 2023-12-25
add relay mode

maltee1:
I currently don't have time for thorough testing & cleanup, but I have been able to relay a message to Signal.
Please give this a whirl if you have time. I will otherwise do thorough testing and cleanup at some later point.

### DIFF
--- a/portal.go
+++ b/portal.go
@@ -1895,7 +1895,7 @@ func (portal *Portal) addRelaybotFormat(userID id.UserID, content *event.Message
 	content.EnsureHasHTML()
 	data, err := portal.bridge.Config.Bridge.Relay.FormatMessage(content, userID, *member)
 	if err != nil {
-		portal.log.Error().Msgf("Failed to apply relaybot format: %s", err)
+		portal.log.Err(err).Msg("Failed to apply relaybot format")
 	}
 	content.FormattedBody = data
 	return true
--- a/portal.go
+++ b/portal.go
@@ -85,6 +85,8 @@ type Portal struct {
 	currentlyTypingLock sync.Mutex
 
 	latestReadTimestamp uint64 // Cache the latest read timestamp to avoid unnecessary read receipts
+
+	relayUser *User
 }
 
 const recentMessageBufferSize = 32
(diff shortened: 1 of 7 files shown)


maltee1:
I would request a review at this point.

One major problem is that the bridge doesn't seem to support outgoing captions yet (although the code looks like it does). That means that Images are sent without "MatrixUser sent an Image" etc. and thus cannot be assigned to a user. I sent an image with a caption from Telegram via double-puppeting from Telegram, so the format should be correct. I currently don't know how to fix that but I will file a bug.

Text messages and emotes work fine.

portal.go
```diff
@@ -1824,3 +1867,21 @@ func (portal *Portal) HandleNewDisappearingMessageTime(newTimer uint32) {
...
+
+func (portal *Portal) addRelaybotFormat(userID id.UserID, content *event.MessageEventContent) bool {
+	member := portal.MainIntent().Member(portal.MXID, userID)
+	if member == nil {
+		member = &event.MemberEventContent{}
+	}
+	content.EnsureHasHTML()
+	data, err := portal.bridge.Config.Bridge.Relay.FormatMessage(content, userID, *member)
+	if err != nil {
+		portal.log.Error().Msgf("Failed to apply relaybot format: %s", err)
```
tulir, line 1883: Using `.Msgf` is not allowed (unfortunately we forgot to tell Scott before he wrote the bridge)

```suggestion
		portal.log.Err(err).Msg("Failed to apply relaybot format")
```

(loggers should also be passed through contexts with fancy context parameters, like event id and stuff in event handlers, but that needs more refactoring)

tulir (approved):

portal.go
```diff
@@ -1824,3 +1867,21 @@ func (portal *Portal) HandleNewDisappearingMessageTime(newTimer uint32) {
...
+func (portal *Portal) HasRelaybot() bool {
+	return portal.bridge.Config.Bridge.Relay.Enabled && len(portal.RelayUserID) > 0
+}
+
+func (portal *Portal) addRelaybotFormat(userID id.UserID, content *event.MessageEventContent) bool {
+	member := portal.MainIntent().Member(portal.MXID, userID)
+	if member == nil {
+		member = &event.MemberEventContent{}
+	}
+	content.EnsureHasHTML()
```
tulir, line 1880: To fix captions after rebasing on main

```suggestion
	if content.FileName == "" {
		content.FileName = content.Body
	}
	content.EnsureHasHTML()
```

PR #399, opened 2023-12-27
modules: un-modulify libsignalgo and signalmeow

### LINKED ISSUE #398: go mod tidy fails
Running `go mod tidy` fails on `main`.

```
$ go mod tidy
go: finding module for package go.mau.fi/mautrix-signal/pkg/signalmeow/protobuf
go: finding module for package go.mau.fi/mautrix-signal/pkg/signalmeow
go.mau.fi/mautrix-signal imports
        go.mau.fi/mautrix-signal/pkg/signalmeow: cannot find module providing package go.mau.fi/mautrix-signal/pkg/signalmeow: unrecognized import path "go.mau.fi/mautrix-signal/pkg/signalmeow": reading https://go.mau.fi/mautrix-signal/pkg/signalmeow?go-get=1: 404 Not Found
go.mau.fi/mautrix-signal imports
        go.mau.fi/mautrix-signal/pkg/signalmeow/protobuf: cannot find module providing package go.mau.fi/mautrix-signal/pkg/signalmeow/protobuf: unrecognized import path "go.mau.fi/mautrix-signal/pkg/signalmeow/protobuf": reading https://go.mau.fi/mautrix-signal/pkg/signalmeow/protobuf?go-get=1: 404 Not Found
```

There's a go.mod and go.sum in pkg/signalmeow, which I think is the cause of
this.

sumnerevans:
I don't see any reason why they should have their own go.mod. I think
this is legacy from when I had them in separate repositories.

If we ever need to split those packages out into their own repo, we can
just:

1. Copy the directories to a new repo
2. Copy the go.mod/go.sum files
3. Run `go mod tidy` in this repo and the new repo.

This commit also removes the go.work and go.work.sum files.

Closes #398


tulir (approved):
The reason was to allow cleaner use of signalmeow from other things, but go workspaces are somewhat annoying (tidy breaks as mentioned), so maybe this is better for now

tulir:
Oh right, this will break tests in CI because libsignalgo has a broken test right now. Also need to merge the two steps in actions (adding `-ldflags "-extldflags -lm"` to the "Run bridge tests" entry and removing "Run libsignalgo tests")

tulir:
https://github.com/mautrix/signal/commit/175fde6a940001c5e4f9ebeae68e38bc646e379e should fix some of the broken tests, but `TestSessionCipherWithBadStore` and `TestSealedSenderSession` are still broken

PR #403, opened 2023-12-30
provisioning: cleanup

sumnerevans:
* use libserv
* use proper context key
* fix resolve identifier response fields

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module go.mau.fi/mautrix-signal
 go 1.20
 
 require (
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -1,8 +1,11 @@
 module go.mau.fi/mautrix-signal
 
-go 1.20
+go 1.21
```
tulir, line 3: Minimum Go version should only be bumped when the previous one is EOL

tulir (approved):

PR #412, opened 2024-01-03
pre-commit: use Stringer(x, y) intsead of Str(x, y.String())

sumnerevans:

### DIFF
--- a/portal.go
+++ b/portal.go
@@ -307,7 +307,7 @@ func (portal *Portal) handleMatrixMessages(msg portalMatrixMessage) {
 	case event.EventReaction:
 		portal.handleMatrixReaction(ctx, msg.user, msg.evt)
 	default:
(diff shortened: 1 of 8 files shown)


portal.go
```diff
@@ -307,7 +307,7 @@ func (portal *Portal) handleMatrixMessages(msg portalMatrixMessage) {
 	case event.EventReaction:
 		portal.handleMatrixReaction(ctx, msg.user, msg.evt)
 	default:
-		log.Warn().Str("type", msg.evt.Type.String()).Msg("Unhandled matrix message type")
+		log.Warn().Stringer("type", &msg.evt.Type).Msg("Unhandled matrix message type")
```
tulir, line 310: Not sure if taking a pointer is nicer than accessing the type directly 🤔

```suggestion
		log.Warn().Str("type", msg.evt.Type.Type).Msg("Unhandled matrix message type")
```

tulir (approved):

PR #417, opened 2024-01-06
treewide: fix all ignored errors

sumnerevans:

### DIFF
--- a/pkg/signalmeow/groups.go
+++ b/pkg/signalmeow/groups.go
@@ -173,10 +173,9 @@ func (cli *Client) GetAuthorizationForToday(ctx context.Context, masterKey libsi
 		zlog.Err(err).Msg("DeriveGroupSecretParamsFromMasterKey error")
 		return nil, err
 	}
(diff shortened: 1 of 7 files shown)


pkg/signalmeow/groups.go
```diff
@@ -174,6 +174,9 @@ func (cli *Client) GetAuthorizationForToday(ctx context.Context, masterKey libsi
 		return nil, err
 	}
 	randomness, err := libsignalgo.GenerateRandomness()
+	if err != nil {
+		return nil, fmt.Errorf("unable to read randomness: %w", err)
```
tulir, line 178: GenerateRandomness should probably just be changed to panic like all other random generating functions 🤔

tulir (approved):

pkg/signalmeow/groups.go
sumnerevans: right, I guess if there's not enough randomness, things have gone horribly wrong

PR #418, opened 2024-01-07
staticcheck: fix all issues

sumnerevans:


pkg/signalmeow/profile.go
```diff
@@ -300,22 +299,6 @@ func decryptString(key *libsignalgo.ProfileKey, encryptedText []byte) (string, e
 	return string(data), err
 }
 
-func encryptString(key libsignalgo.ProfileKey, plaintext string, paddedLength int) ([]byte, error) {
```
tulir, line 303: Probably shouldn't remove this

sumnerevans: why was it not used?

tulir: It'll be needed when we want to modify profiles, but we don't do that yet

PR #428, opened 2024-01-25
Restore "whoami" for provisioning API

AndrewFerr:


provisioning.go
```diff
@@ -388,6 +406,47 @@ func (prov *ProvisioningAPI) checkSessionAndReturnHandle(ctx context.Context, w
...
+		if puppet == nil {
+			status = http.StatusInternalServerError
+			data.Signal = &WhoAmIResponseSignal{
+				Number: user.SignalUsername,
+				Ok:     false,
+				Error:  "Failed to get puppet by Signal ID",
+			}
+			log.Error().
+				Str("signal_id", user.SignalID.String()).
+				Msg(data.Signal.Error)
```
tulir, line 433: This case doesn't really do anything, so it should just be removed. The `ok` boolean could also be removed or renamed. Instead there should be a boolean indicating if the client is connected to signal servers

provisioning.go
```diff
@@ -388,6 +406,47 @@ func (prov *ProvisioningAPI) checkSessionAndReturnHandle(ctx context.Context, w
...
+				UUID:   user.SignalID.String(),
+				Name:   puppet.Name,
+				Ok:     true,
+			}
+		}
+	}
+	jsonResponse(w, status, Response{
+		Success:        true,
+		Status:         "ok",
+		WhoAmIResponse: data,
```
tulir, line 446: The `Response` struct shouldn't be used for anything other than login (resolve identifier was just implemented wrong)

provisioning.go
AndrewFerr: `ok` was returned in the signald version of the bridge, so IMO it should be kept around one way or another to preserve backwards compatibility.

I changed `ok` to contain the connection status; if you'd rather that be in a more aptly-named property (like `connected` or something), I'll change it.

tulir (approved):

PR #429, opened 2024-01-25
Deduplicate users' UUIDs during schema upgrade

AndrewFerr:
Otherwise, upgrading to v16 may hit conflicts if two users share the same Signal UUID, which was possible in prior DB versions.

---

This is useful only when upgrading a DB from the signald version of the bridge, where some Matrix users had logged into the same Signal account (and thus shared the same UUID in the "users" table).

Though this removes Signal UUID associations from Matrix users, migrating from the signald version requires re-logging in anyways.

The only time this would not be safe is when migrating a DB that had already been migrated from signald (i.e. a v13-v15 version) and users logged back in, with some sharing the same Signal UUID.


tulir (approved):
Hmm, I guess the DM portal receiver update happens before this point, so it's fine to clear the user rows 🤔

PR #430, opened 2024-01-25
Ensure user is invited to portals on update

AndrewFerr:
because if a user ever leaves a portal, there is no obvious way for them to rejoin it.

This restores 468f655, which was lost during a refactor.


tulir (approved):
Originally it was very inefficient, but now it might be fine since ensureInvited always checks the cache

PR #431, opened 2024-01-25
Don't replace bridged reactions sent from Matrix

AndrewFerr:
This is the second commit of https://github.com/mautrix/signalgo/pull/20 that I had [mentioned](https://github.com/mautrix/signalgo/pull/20#issuecomment-1843868011) wanting to open a PR for.

### DIFF
--- a/portal.go
+++ b/portal.go
@@ -940,6 +940,12 @@ func (portal *Portal) handleSignalReaction(sender *Puppet, react *signalpb.DataM
 	}
 	intent := sender.IntentFor(portal)
 	if existingReaction != nil {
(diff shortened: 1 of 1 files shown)


sumnerevans:
In https://github.com/mautrix/signalgo/pull/20#issuecomment-1825854788:

> Ah, the error I'm getting is `M_DUPLICATE_ANNOTATION (HTTP 400): Can't send same reaction twice`, when a bridged Matrix user sends a reaction in a portal. This happens because the attempt to delete the original reaction fails with `M_FORBIDDEN (HTTP 403): You don't have permission to redact events`.

This seems like the actual issue, we should just make the redaction work properly instead of patching it with this hack.

PR #434, opened 2024-01-29
Add config to deny `login-matrix`

AndrewFerr:
This changes user bridge permissions to be more like mautrix-telegram's, which has a separate permission level for controlling access to double-puppeting.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -29,6 +29,7 @@ import (
 	"maunium.net/go/mautrix/event"
 	"maunium.net/go/mautrix/id"
 
+	"go.mau.fi/mautrix-signal/config"
 	"go.mau.fi/mautrix-signal/pkg/signalmeow"
 )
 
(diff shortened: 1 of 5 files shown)


AndrewFerr:
Note that this changes the `"user"` permission level to disallow double-puppeting, so unless some kind of migration step is added to this PR, existing bridges that wish to not disable double-puppeting for users who have enabled it must either the `"user"` level in their config to `"full"`.

AndrewFerr:
If this is a useful change, it can be moved into mautrix-go.

tulir:
Probably needs to be on the mautrix-go side, no point in adding special cases like that here since all the bridges have double puppeting. Changing the definition of permission names also isn't an option, so will need to make up new names. I might try to reinvent the permission system later to better enable adding new permissions, like maybe having a list of permissions instead of a linear level

AndrewFerr:
mautrix-go PR: https://github.com/mautrix/go/pull/170

PR #437, opened 2024-02-02
Add config to disallow manual double puppeting

AndrewFerr:
Depends on https://github.com/mautrix/go/pull/174

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -18,7 +18,6 @@ package config
 
 import (
 	"maunium.net/go/mautrix/bridge/bridgeconfig"
-	"maunium.net/go/mautrix/id"
 )
 
 type Config struct {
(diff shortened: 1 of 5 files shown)


olmari:
I have no direct opinion for this kind of option to exist in some form, but one thing I like to point out:

> # Relevant for users on servers where automatic double puppeting has not been enabled.

For example we use automatic double puppeting for users in our server, but we also allow our users to use other HS if they like to (usually their own) and still allow usage of our bridge, hence manual and automatic DP is not mutually exclusive as a whole in our case.

AndrewFerr:
Note this has been merged into the Element fork as of [v0.5.0-mod-1](https://github.com/element-hq/mautrix-signal/releases/tag/v0.5.0-mod-1).

PR #440, opened 2024-02-06
Satisfy static checks

AndrewFerr:
This allows the `staticcheck` tool to report no issues.

### DIFF
--- a/main.go
+++ b/main.go
@@ -114,7 +114,7 @@ func (br *SignalBridge) Init() {
 		Bridge: br,
 	}
 
-	br.Metrics = NewMetricsHandler(br.Config.Metrics.Listen, br.Log.Sub("Metrics"), br.DB)
+	br.Metrics = NewMetricsHandler(br.Config.Metrics.Listen, br.ZLog.With().Str("component", "metrics").Logger(), br.DB)
 	br.MatrixHandler.TrackEventDuration = br.Metrics.TrackMatrixEvent
 
 	signalFormatParams = &signalfmt.FormatParams{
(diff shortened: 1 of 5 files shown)


metrics.go
```diff
@@ -276,9 +276,15 @@ func (mh *MetricsHandler) updateStats() {
...
 func (mh *MetricsHandler) startUpdatingStats() {
 	defer func() {
-		err := recover()
-		if err != nil {
-			mh.log.Fatalfln("Panic in metric updater: %v\n%s", err, string(debug.Stack()))
+		r := recover()
+		if r != nil {
+			evt := mh.log.Fatal()
+			if err, ok := r.(error); ok {
+				evt = evt.Stack().Err(err)
```
tulir, line 283: I don't think `.Stack()` actually does anything, need to always use `debug.Stack()`

main.go
```diff
@@ -197,7 +197,7 @@ func (br *SignalBridge) Start() {
 func (br *SignalBridge) Stop() {
 	br.Metrics.Stop()
 	for _, user := range br.usersByMXID {
-		br.Log.Debugln("Disconnecting", user.MXID)
+		br.ZLog.Debug().Stringer("user_id", user.MXID).Msg("Disconnecting")
```
tulir, line 200:
```suggestion
		br.ZLog.Debug().Stringer("user_id", user.MXID).Msg("Disconnecting user")
```

tulir (approved):

metrics.go
AndrewFerr: Gah, you're right -- it only works with `pkg/errors`.

sumnerevans (approved):

Issue #396, opened 2023-12-26
Don't "downgrade" displaynames


maltee1:
The go bridge seems to be suffering from the same problem the python bridge suffered from: renaming of users on every sync. For now I can only see users being renamed back and forth between phone number and profile name, so a good first measure might be to not downgrade a name, i.e. not rename to a phone number if the current one is a profile name.
I can't yet tell whether some profile name information will eventually lose sync between different bridge users, I guess I'll wait and file another bug report in that case.

maltee1:
I think this was solved with "name quality" in the python bridge, but I think a more straightforward way would be to just store the profile name and check if it's empty - in that case, just use the previously known profile name.

tulir:
Name quality is probably required, possibly even more complicated than the python bridge. It should include the profile fetch timestamp or some other version identifier for the profile name.

maltee1:
This appears to affect avatars, too. While re-setting the same avatar seems fixed, I'm still seeing avatars being removed and re-set.

Is it possible that contact info is sometimes empty and not updating empty contact info would already solve the problem? Maybe checking for empty profileKey or something would work?

AndrewFerr:
> Is it possible that contact info is sometimes empty and not updating empty contact info would already solve the problem? Maybe checking for empty profileKey or something would work?

What can happen is that the contact info isn't _entirely_ empty, but fields that had been filled on an earlier fetch are empty on a later fetch.  For example, a fetched info can have all but its phone number be empty, even if an earlier fetch included a profile name.

maltee1:
I figured out the issue for profile names in my particular case I think. It seems the bridge attempts to sync profiles regardless of whether a profile key is available or not (`profile_key == \x` in the database). If a bridge user has someone in their contact list but hasn't spoken to them, their profile will return empty. If someone else has their profile key and can therefore get their profile name, the bridge will toggle between name and phone number.

It probably doesn't make sense to attempt fetching profiles if no profile key is available. Disabling that should significantly improve the situation.

PR #445, opened 2024-02-08
Always fetch profile on update info, and reduce profile data leaks

AndrewFerr:
This helps with #396 insofar as it should prevent puppets' displaynames/avatars from being set/unset as often.  However, it still allows displaynames to be "downgraded" to avoid sharing profile info with users that it's not shared with in Signal.

It's worth mentioning that until Matrix supports locally-overrideable profile info like Signal does, any attempt to lock down bridged profile visibility to the same extent that Signal does will either be overly strict, or not strict enough.

See commit descriptions for details.

### DIFF
--- a/pkg/signalmeow/profile.go
+++ b/pkg/signalmeow/profile.go
@@ -162,6 +162,10 @@ func (cli *Client) RetrieveProfileByID(ctx context.Context, signalID uuid.UUID) 
 	return profile, nil
 }
 
+func (cli *Client) IsProfileHiddenFromOthers(ctx context.Context, profileUUID uuid.UUID) (bool, error) {
+	return cli.Store.ContactStore.IsProfileHiddenFromOthers(ctx, profileUUID)
+}
+
 func (cli *Client) fetchProfileByID(ctx context.Context, signalID uuid.UUID) (*Profile, error) {
 	log := zerolog.Ctx(ctx)
 	profileKey, err := cli.ProfileKeyForSignalID(ctx, signalID)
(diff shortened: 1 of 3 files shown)


tulir:
The point of not fetching was to avoid spamming requests on startup, but I can see how that would cause issues if there are multiple users and part of them haven't fetched the profile yet 🤔 I guess passing `info` to `UpdateInfo` can be removed entirely, and later `ContactByID` can be updated to persist the last fetch time in the database so it wouldn't refetch on every startup.

Not sure about the other commit

AndrewFerr:
How should the fetch timestamps be used?  Is the idea to re-fetch only if the previous fetch is older than some fixed amount of time?

tulir:
Yeah something like that. I guess another option would be to not request contacts on every startup, or only fetch profile if something in the contact list changes.

Anyway, removed the info parameter for now https://github.com/mautrix/signal/commit/2109782afdbda3700ca292f98282e437cb45075d

AndrewFerr:
Perhaps the attempt to reduce profile data leaks isn't worth the trouble, as it's all too easy for profiles to be leaked by regular Matrix behaviour, like inviting someone to a room with a Signal puppet in it, or to simply run the bridge on a server with a searchable user directory.

Also, since (AIUI) the old version of the bridge didn't apply the kind of profile privacy done in this PR, it wouldn't be a regression to abandon it.

In light of that, I plan to write an alternate PR that instead tries to use the most up-to-date profile data possible.

AndrewFerr:
Replacing with #449.

PR #448, opened 2024-02-13
Update group info

maltee1:
This seems to do what it's supposed to do, although I could probably test more extensively

### DIFF
--- a/portal.go
+++ b/portal.go
@@ -2242,7 +2242,10 @@ func (portal *Portal) SyncParticipants(ctx context.Context, source *User, info *
 			delete(currentMembers, mxid)
 		}
 	}
-	for mxid, _ := range currentMembers {
+	for mxid, membership := range currentMembers {
+		if membership == event.MembershipLeave {
+			continue
+		}
 		user := portal.bridge.GetUserByMXIDIfExists(mxid)
 		if user != nil {
 			if user.IsLoggedIn() {
--- a/portal.go
+++ b/portal.go
@@ -2131,21 +2131,27 @@ func (portal *Portal) SyncParticipants(ctx context.Context, source *User, info *
 		if member.UserID != source.SignalID && portal.MXID != "" {
 			userIDs[intent.UserID] = ((int)(member.Role) >> 1) * 50
 		}
+		delete(currentMembers, intent.UserID)
 		if portal.MXID != "" {
-			err := intent.EnsureJoined(ctx, portal.MXID)
-			if err != nil {
-				log.Err(err).Stringer("signal_user_id", member.UserID).Msg("Failed to ensure user is joined to portal")
+			if currentMembers[intent.UserID] != event.MembershipJoin {
+				err := intent.EnsureJoined(ctx, portal.MXID)
+				if err != nil {
+					log.Err(err).Stringer("signal_user_id", member.UserID).Msg("Failed to ensure user is joined to portal")
+				}
 			}
 			if puppet.customIntent == nil {
 				user := portal.bridge.GetUserBySignalID(member.UserID)
 				if user != nil {
-					user.ensureInvited(ctx, intent, portal.MXID, false)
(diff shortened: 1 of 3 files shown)


commands.go
```diff
@@ -642,3 +643,22 @@ func fnCleanupLostPortals(ce *WrappedCommandEvent) {
...
+		Description: "Sync this portal.",
+	},
+	RequiresPortal: true,
+	RequiresLogin:  true,
+	RequiresAdmin:  true,
+}
+
+func fnSyncPortal(ce *WrappedCommandEvent) {
+	info, _ := ce.User.Client.RetrieveGroupByID(ce.Ctx, ce.Portal.GroupID(), ce.Portal.Revision)
+	if info != nil {
```
tulir, line 661: This should log errors and reply if it fails or succeeds

commands.go
```diff
@@ -642,3 +643,22 @@ func fnCleanupLostPortals(ce *WrappedCommandEvent) {
...
+var cmdSyncPortal = &commands.FullHandler{
+	Func: wrapCommand(fnSyncPortal),
+	Name: "sync-portal",
+	Help: commands.HelpMeta{
+		Section:     HelpSectionPortalManagement,
+		Description: "Sync this portal.",
+	},
+	RequiresPortal: true,
+	RequiresLogin:  true,
+	RequiresAdmin:  true,
```
tulir, line 656: Probably doesn't need to require admin

portal.go
```diff
@@ -1838,6 +1841,72 @@ func (portal *Portal) UpdateDMInfo(ctx context.Context, forceSave bool) {
 	}
 }
 
+func (portal *Portal) updatePowerLevelsAndJoinRule(ctx context.Context, info *signalmeow.Group, members map[id.UserID]int) {
+	log := zerolog.Ctx(ctx).With().
+		Str("function", "updatePowerLevelsAndJoinRule").
+		Logger()
+	log.Trace().Msg("Updating power levels and join rule")
+	state, err := portal.MainIntent().State(ctx, portal.MXID)
+	if err != nil {
```
tulir, line 1850: This shouldn't fetch the whole state if it just needs 2 events

portal.go
```diff
@@ -2040,29 +2109,167 @@ func (portal *Portal) updateAvatarInRoom(ctx context.Context, sender *Puppet) {
...
+		membership := currentMembers[mxid]
+		var err error
+		if membership == event.MembershipJoin || membership == event.MembershipBan {
+			_, err = portal.MainIntent().SendCustomMembershipEvent(ctx, portal.MXID, mxid, event.MembershipLeave, "")
+		}
+		if membership != event.MembershipInvite {
+			_, err = portal.MainIntent().SendCustomMembershipEvent(ctx, portal.MXID, mxid, event.MembershipInvite, "")
+		}
+		if err != nil {
+			log.Warn().Stringer("mxid", mxid).Msg("Couldn't change membership to invite")
```
tulir, line 2175: Missing errors in most of the logs here. The errors for the two cases should probably also be logged separately.

portal.go
```diff
@@ -2040,29 +2109,167 @@ func (portal *Portal) updateAvatarInRoom(ctx context.Context, sender *Puppet) {
...
+			}
+			delete(currentMembers, mxid)
+		}
+	}
+	for mxid, membership := range currentMembers {
+		if membership == event.MembershipLeave {
+			continue
+		}
+		user := portal.bridge.GetUserByMXIDIfExists(mxid)
+		if user != nil {
```
tulir, line 2256: When getting by mxid it's more efficient to try puppet first and then user, because determining if a mxid is a bridge ghost doesn't require a database query

commands.go
maltee1: I didn't intend the sync-portal command for merge, so I didn't bother to make it pretty. It should probably be integrated in a more general `sync` command at some point, I just needed a way to trigger the sync of a single portal for testing.

tulir (approved):

PR #449, opened 2024-02-14
Use latest profile info

AndrewFerr:
This is an alternative approach to #445 for helping with #396: if a Signal profile cannot be fetched by a particular bridged user's Signal account, use the profile info from the most recent successful fetch done by some other bridge user.

Profiles retrieved this way are used only as a fallback for setting a Signal puppet's displayname/avatar, not for updating the contact store.

### DIFF
--- a/config/bridge.go
+++ b/config/bridge.go
@@ -169,12 +169,12 @@ type DisplaynameParams struct {
 func (bc BridgeConfig) FormatDisplayname(contact *types.Contact) string {
 	var buffer strings.Builder
 	_ = bc.displaynameTemplate.Execute(&buffer, DisplaynameParams{
-		ProfileName: contact.ProfileName,
+		ProfileName: contact.Profile.Name,
 		ContactName: contact.ContactName,
 		//Username:    contact.Username,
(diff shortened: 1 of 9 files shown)


tulir:
Querying the database for profiles fetched by other users seems weird 🤔

The mechanism used in other bridges and the old signal bridge is the name quality field: if fetching doesn't return profile info (or it fails) and the puppet already has a profile name, then just don't try to update the name at all.

AndrewFerr:
I used profile recency as a simpler alternative to a proper quality metric, which I wasn't sure how to implement properly in light of [it being more complicated in this version of the bridge](https://github.com/mautrix/signal/issues/396#issuecomment-1870046745).

The idea here is that someone else having a more recent fetch for a given user's profile should be analogous to higher-quality profile data for that user.

This is admittedly an incomplete solution to the name quality problem, so feel free to close this if you'd rather go straight to a full solution.

tulir:
It might not necessarily need to be more complicated since the profile info is refetched each time, at least assuming a successful fetch means it's up to date. I'll see if I can come up with a simple solution

Issue #444, opened 2024-02-07
Migration fails on user_portal_portal_fkey constraint


KitsuneRal:
Here's the relevant piece in the logs:
```
Feb 07 11:50:02 clup matrix-postgres[3699196]: 2024-02-07 10:50:02.366 UTC [360111] ERROR:  foreign key constraint "user_portal_portal_fkey" cannot be implemented
Feb 07 11:50:02 clup matrix-postgres[3699196]: 2024-02-07 10:50:02.366 UTC [360111] DETAIL:  Key columns "portal_receiver" and "receiver" are of incompatible types: uuid and text.
Feb 07 11:50:02 clup matrix-postgres[3699196]: 2024-02-07 10:50:02.366 UTC [360111] STATEMENT:  ALTER TABLE "user" ADD COLUMN space_room TEXT;
Feb 07 11:50:02 clup matrix-postgres[3699196]:         
Feb 07 11:50:02 clup matrix-postgres[3699196]:         DROP TABLE IF EXISTS user_portal;
Feb 07 11:50:02 clup matrix-postgres[3699196]:         CREATE TABLE user_portal (
Feb 07 11:50:02 clup matrix-postgres[3699196]:             user_mxid       TEXT,
Feb 07 11:50:02 clup matrix-postgres[3699196]:             portal_chat_id  TEXT,
Feb 07 11:50:02 clup matrix-postgres[3699196]:             portal_receiver uuid,
Feb 07 11:50:02 clup matrix-postgres[3699196]:             last_read_ts    BIGINT  NOT NULL DEFAULT 0,
Feb 07 11:50:02 clup matrix-postgres[3699196]:             in_space        BOOLEAN NOT NULL DEFAULT false,
Feb 07 11:50:02 clup matrix-postgres[3699196]:         
Feb 07 11:50:02 clup matrix-postgres[3699196]:             PRIMARY KEY (user_mxid, portal_chat_id, portal_receiver),
Feb 07 11:50:02 clup matrix-postgres[3699196]:             CONSTRAINT user_portal_user_fkey FOREIGN KEY (user_mxid)
Feb 07 11:50:02 clup matrix-postgres[3699196]:                 REFERENCES "user"(mxid) ON UPDATE CASCADE ON DELETE CASCADE,
Feb 07 11:50:02 clup matrix-postgres[3699196]:             CONSTRAINT user_portal_portal_fkey FOREIGN KEY (portal_chat_id, portal_receiver)
[...]

Issue #425, opened 2024-01-19
Migration does not clean up portals without users properly


verymilan:
[matrix discussion a](https://matrix.to/#/!mautrix-signal-v6:maunium.net/$1lKrX2CX43Jkk25CZMJ-eRyO_T5rdNzxi-ipXPwtHuw?via=maunium.net&via=matrix.org&via=tchncs.de) / [matrix discussion b](https://matrix.to/#/!mautrix-signal-v6:maunium.net/$z-LbKHssNbFqYjgjqrtmfJNLiElJzpHXKchUaVXGK0Y?via=maunium.net&via=matrix.org&via=tchncs.de)

Right now, coming from the latest release to the rewrite on the latest commit, the migration process fails with the following error:

`{"level":"fatal","error":"failed to run upgrade #15: pq: null value in column \"receiver\" of relation \"portal\" violates not-null constraint","db_section":"main","pq_error":{"severity":"ERROR","code":"not_null_violation","detail":"Failing row contains (b8c23d3a-85ca-443d-b275-6bcf0cd8009f, null, null, +49 176 redacted (Signal), null, f, null, null, f, f, 0, null, null).","schema":"public","table":"portal","column":"receiver","file":"execMain.c","line":"1968","routine":"ExecConstraints"},"time":"2024-01-18T20:35:46.541347008Z","message":"Failed to initialize database"}`

While the portal row looks like

```
               chat_id                |   receiver    | mxid |           name            | topic | encrypted | avatar_hash | avatar_url | name_set | avatar_set | revision | expiration_time | relay_user_id
--------------------------------------+---------------+------+---------------------------+-------+-----------+-------------+------------+----------+------------+----------+-----------------+---------------
 b8c23d3a-85ca-443d-b275-6bcf0cd8009f | +49175redacted |      | +49 176 redacted (Signal) |       | f         |             |            | f        | f          |        0 |                 |
(1 row)
```

and `SELECT * FROM "user" WHERE username='+49175redacted'` returns 0 rows.

When deleting that row above, it just jumps to the next unwelcome row.

PR #451, opened 2024-02-15
DB migration fixes

AndrewFerr:
This fixes some potential failures on migrating the `main` store to v16.

### DIFF
--- a/database/upgrades/16-refactor-postgres.sql
+++ b/database/upgrades/16-refactor-postgres.sql
@@ -51,7 +51,7 @@ INSERT INTO lost_portals SELECT mxid, chat_id, receiver FROM portal WHERE mxid<>
 UPDATE portal SET mxid=NULL WHERE mxid='';
 ALTER TABLE portal ADD CONSTRAINT portal_mxid_unique UNIQUE(mxid);
 -- Delete any portals that aren't associated with logged-in users.
-DELETE FROM portal WHERE receiver<>'' AND receiver NOT IN (SELECT username FROM "user" WHERE uuid IS NOT NULL);
+DELETE FROM portal WHERE receiver<>'' AND receiver NOT IN (SELECT username FROM "user" WHERE username IS NOT NULL AND uuid IS NOT NULL);
 -- Change receiver to uuid instead of phone number, also add nil uuid for groups.
 UPDATE portal SET receiver=(SELECT uuid FROM "user" WHERE username=receiver AND uuid IS NOT NULL LIMIT 1) WHERE receiver<>'';
 UPDATE portal SET receiver='00000000-0000-0000-0000-000000000000' WHERE receiver='';
--- a/database/upgrades/16-refactor-postgres.sql
+++ b/database/upgrades/16-refactor-postgres.sql
@@ -18,6 +18,9 @@ ALTER TABLE message ALTER COLUMN part_index SET NOT NULL;
 ALTER TABLE reaction ADD COLUMN _part_index INTEGER NOT NULL DEFAULT 0;
 
 -- Re-add the dropped constraints (but with part index and no chat)
+DELETE FROM message
+    WHERE (sender, timestamp, part_index, signal_receiver)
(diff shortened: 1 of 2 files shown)


database/upgrades/16-refactor-postgres.sql
```diff
@@ -48,7 +51,7 @@ INSERT INTO lost_portals SELECT mxid, chat_id, receiver FROM portal WHERE mxid<>
 UPDATE portal SET mxid=NULL WHERE mxid='';
 ALTER TABLE portal ADD CONSTRAINT portal_mxid_unique UNIQUE(mxid);
 -- Delete any portals that aren't associated with logged-in users.
-DELETE FROM portal WHERE receiver<>'' AND receiver NOT IN (SELECT username FROM "user" WHERE uuid IS NOT NULL);
+DELETE FROM portal WHERE receiver<>'' AND receiver NOT IN (SELECT username FROM "user" WHERE username IS NOT NULL AND uuid IS NOT NULL);
```
AndrewFerr, line 54: This should address #444.  Without it, the `NOT IN` will match no rows if any null username exists, which would prevent any rows from being deleted.  That behaviour of `NOT IN` is explained by https://modern-sql.com/concept/three-valued-logic#not-in-null.

tulir: Do you mean #425?

tulir (approved):
Looks fairly sensible

database/upgrades/16-refactor-postgres.sql
AndrewFerr: Ah, yes

AndrewFerr: Hmm, looks like this can hang...  I'll go back to the drawing board for this

AndrewFerr: This was correct after all, but was triggering some slow CASCADE actions on our big database.

AndrewFerr:
ea9fc29 isn't strictly necessary, but it unblocked an extremely slow migration (>14 hours) on a big database & may come in handy for others.

I also haven't tested d9b139a but assume it's no different from the Postgres case.

tulir:
Cascades aren't slow, it's probably just missing some indexes

AndrewFerr:
> Cascades aren't slow, it's probably just missing some indexes

Alright, that makes more sense.  In any case, I couldn't find the right index to add that made the cascade any faster.  An index would be ideal, though.

database/upgrades/17-refactor-sqlite.sql
```diff
@@ -50,6 +50,10 @@ CREATE TABLE reaction_new (
 );
 
 
+DELETE FROM message
+    WHERE (sender, timestamp, part_index, signal_receiver)
+    IN (SELECT DISTINCT sender, timestamp, part_index, signal_receiver FROM message GROUP BY (sender, timestamp, part_index, signal_receiver) HAVING COUNT(*)>1);
```
tulir, line 55: SQLite doesn't seem to like the parentheses around the group by columns. Also, I think the `DISTINCT` is redundant because it's grouped by the same columns?

database/upgrades/16-refactor-postgres.sql
```diff
@@ -47,14 +50,27 @@ INSERT INTO lost_portals SELECT mxid, chat_id, receiver FROM portal WHERE mxid<>
 -- Make mxid column unique (requires using nulls for missing values)
 UPDATE portal SET mxid=NULL WHERE mxid='';
 ALTER TABLE portal ADD CONSTRAINT portal_mxid_unique UNIQUE(mxid);
+-- Drop foreign keys that make CASCADEs prohibitively slow.
+ALTER TABLE message DROP CONSTRAINT message_portal_fkey;
+ALTER TABLE reaction DROP CONSTRAINT reaction_message_fkey;
```
tulir, line 55: It could probably just skip re-adding the constraints 25 lines above rather than delete them here

PR #461, opened 2024-02-22
matrix -> signal group metadata

maltee1:
This still needs conflict resolution and I think groupChanges are pushed to pending members as well, which depends on #448 .


maltee1:
depends on https://github.com/mautrix/go/pull/193 (I didn't update the dependency)

PR #464, opened 2024-02-24
Prevent downgrading ghost user info by default

### LINKED ISSUE #396: Don't "downgrade" displaynames

tulir:
Fixes #396
Closes #449


tulir:
@AndrewFerr I think this should solve the issue by simply not updating info if the puppet already had profile info and profile fetching fails, which feels less weird than fetching profiles from another user's database

Issue #466, opened 2024-02-28
Error "cannot convert slice" while relaying message


LukasNeuberger:
I am using the bridge docker image 19929f6/commit 2717c53 to relay messages between Whatsapp and Signal groups.
While relaying a message from a Whatsapp user to Signal I got the following error:
```
panic: runtime error: cannot convert slice with length 0 to array or pointer to array with length 65

goroutine 19480 [running]:
go.mau.fi/mautrix-signal/pkg/signalmeow.decryptPendingMember({_, _}, _, {0x0, 0x13, 0x9c, 0xa, 0x6b, 0xc6, 0x26, ...})
	/builds/mautrix/signal/pkg/signalmeow/groups.go:968 +0x459
go.mau.fi/mautrix-signal/pkg/signalmeow.decryptGroup({0x19a40d8, 0xc000437410}, 0xc0004c1400, {0xc0007088a0, 0x2c})
	/builds/mautrix/signal/pkg/signalmeow/groups.go:394 +0x94b
go.mau.fi/mautrix-signal/pkg/signalmeow.(*Client).fetchGroupByID(0xc0001128c0, {0x19a40d8, 0xc000437410}, {0xc00036d6b0, 0x2c})
	/builds/mautrix/signal/pkg/signalmeow/groups.go:525 +0x3b6
go.mau.fi/mautrix-signal/pkg/signalmeow.(*Client).RetrieveGroupByID(0xc0001128c0, {0x19a40d8, 0xc000437410}, {0xc00036d6b0, 0x2c}, 0x0)
	/builds/mautrix/signal/pkg/signalmeow/groups.go:590 +0x1a9
go.mau.fi/mautrix-signal/pkg/signalmeow.(*Client).SendGroupMessage(0xc0001128c0, {0x19a40d8, 0xc000437380}, {0xc00036d6b0, 0x2c}, 0xc000300d20)
	/builds/mautrix/signal/pkg/signalmeow/sending.go:520 +0x365
main.(*Portal).sendSignalMessage(0xc000118dd0, {0x19a40d8, 0xc0004370b0}, 0xc000300d20, 0xc0000009c0, {0xc000708030, 0x2c})
	/builds/mautrix/signal/portal.go:690 +0x50f
main.(*Portal).handleMatrixMessage(0xc000118dd0, {0x19a40d8, 0xc000436e40}, 0xc0002da540, 0xc0004b01a0)
	/builds/mautrix/signal/portal.go:493 +0xda5
[...]

Issue #465, opened 2024-02-26
`addPendingMembers` does not work properly yet


maltee1:
`addPendingMembers` got me:
```
{"level":"error","chat_id":"RKRZzOb77EydQmUe350+d7kPd9hjq8kUvXDLgE/OmZU=","room_id":"!KTzSYwyBubPkMjwnhi:maltee.de","action":"handle signal group change","sender_uuid":"redacted1","change_ts":1708865287784,"new_revision":5,"action":"decrypt group change","error":"invalid UUID length","time":"2024-02-25T13:48:06.463002512+01:00","message":"DecryptUUID UserId error"}
```
Not sure if this is related to the "profile key not found" error of the invitee a bit later in the logs. Maybe we should store profile keys *before* we do anything else with a given member?

The correspondeng `promotePendingMembers` for some reason didn't work either. The groupChange itself doesn't complain about anyhting, I'm getting a warning about getting rate limited and shortly after there's a bunch of `sendContent too many retries` with the invitee as a sender, those seem to be delivery receipts. But other than that it kind of looks like `promotePendingMembers` is emtpy, because there is no error and `ensureJoined` should work.

I should probably add better error messages to `DecryptGroupChange` and `handleSignalGroupChange`.

Does "invalid UUID length" indicate that decryption went wrong? Or could that be due to some other problem? Wouldn't that mean that the decryption parameters are wrong? Which is odd, because the group master key is taken directly from that group change, so it should be correct.

PR #467, opened 2024-02-29
try fix pendingMember issuse

maltee1:
I don't have time to test this today, but it should at least fix the panic due to a zero length profile key.

I still don't entirely know what's going on, but my understanding is that if a serviceid has 16 bits, it's an ACI, if it's 17 bits and has a zero prepended, it's an ACI, if it's 17 bits and has a 1 prepended, it's a PNI. We don't use PNIs a lot, so I'm just discarding PNIs for now. I couldn't find the stripping of the first byte anywhere in the signal app code, I just assumed it's required (because otherwise the size is wrong)

Please test if you're affected by #466  or #465


tulir (approved):
Looks like this won't break anything more at least

PR #472, opened 2024-03-07
Update profile avatar paths & hashes properly

AndrewFerr:
This is a [leftover commit](https://github.com/mautrix/signal/pull/449/commits/38b8f3e63e9a30ced2f8323312d7e4817e133b1f) from https://github.com/mautrix/signal/pull/449 that may still be relevant.

When running the bridge without this change, avatar updates weren't getting saved to the store when I expected them to & new avatars wouldn't appear in Matrix until a bridge restart.

### DIFF
--- a/puppet.go
+++ b/puppet.go
@@ -325,8 +325,8 @@ func (puppet *Puppet) updateAvatar(ctx context.Context, source *User, info *type
 			// TODO what to do? 🤔
 			return false
 		}
-		puppet.AvatarSet = false
 		puppet.AvatarPath = ""
+		puppet.AvatarHash = info.ContactAvatar.Hash
 	} else {
 		if puppet.AvatarPath == info.Profile.AvatarPath && puppet.AvatarSet {
 			return false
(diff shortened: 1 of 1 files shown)


puppet.go
```diff
@@ -354,19 +354,24 @@ func (puppet *Puppet) updateAvatar(ctx context.Context, source *User, info *type
...
+				Str("avatar_hash", newHash).
+				Str("new_avatar_path", puppet.AvatarPath).
+				Msg("Avatar path changed, but hash didn't")
+			// Path changed, but actual avatar didn't
+			return true
+		}
+		puppet.AvatarHash = newHash
+		source.Client.Store.ContactStore.StoreContact(ctx, *info)
+		err = source.Client.Store.ContactStore.StoreContact(ctx, *info)
+		if err != nil {
```
tulir, line 371: I don't really get what this whole PR is changing. It removes an unnecessary hash check for contact avatars, but that shouldn't break anything even if it's there. These StoreContact calls also shouldn't be here, but maybe they're somehow related to the actual fix?

AndrewFerr: This was supposed to fix a discrepancy in how the `Puppet` & `Contact` structs stored the avatar hash, which could cause the puppet-update function to mistakenly think the avatar didn't change when it really did.

However, that seems to have been fixed ever since the `Contact` stopped storing the avatar hash.  This also makes my `StoreContact` call obsolete now, so you're correct in saying it shouldn't be there.

With that said, I'm still seeing avatar updates get missed due to aggressive caching of fetched profile info, but I'll open a different issue for that.

AndrewFerr:
As explained [here](https://github.com/mautrix/signal/pull/472#discussion_r1521545028), the problem this PR was thought to fix is no longer present.  So that makes this PR just an optional performance change.

Other than avoiding an unneeded hash calculation, it also ensures that changes to a puppet's avatar path are saved to the DB when the puppet's profile metadata comes from a Signal profile (not a contact) and the profile avatar path changed but the hash did not.

puppet.go
```diff
@@ -354,19 +354,19 @@ func (puppet *Puppet) updateAvatar(ctx context.Context, source *User, info *type
 			return true
 		}
 		avatarContentType = http.DetectContentType(avatarData)
+		puppet.AvatarPath = info.Profile.AvatarPath
```
AndrewFerr, line 357: This was moved to before the `return true` so that an path change will end up getting saved to the store even when the hash didn't change.

PR #477, opened 2024-03-12
Make sqlite-compatible migration step

AndrewFerr:
The [schema upgrade](https://github.com/mautrix/signal/pull/464/files#diff-12b4f88ba8afff6707cb0c6ed319508ea582dcecd67857e6c6e1d7eff230f5a0) added in #464 uses `ALTER COLUMN`, so it's incompatible with SQLite.

This PR uses the same strategy as the [v16](https://github.com/mautrix/signal/blob/main/database/upgrades/16-refactor-postgres.sql) & [v17](https://github.com/mautrix/signal/blob/main/database/upgrades/17-refactor-sqlite.sql) upgrades for the main store to add a SQLite-compatible version of this migration.

### DIFF
--- a/pkg/signalmeow/store/upgrades/08-resync-schema-449-postgres.sql
+++ b/pkg/signalmeow/store/upgrades/08-resync-schema-449-postgres.sql
@@ -1,4 +1,6 @@
--- v7 -> v8: Migration from https://github.com/mautrix/signal/pull/449 to match the new v8 upgrade
+-- v7 -> v8: Migration from https://github.com/mautrix/signal/pull/449 to match the new v8 upgrade (Postgres)
+-- only: postgres
+
 ALTER TABLE signalmeow_contacts DROP COLUMN profile_avatar_hash;
 ALTER TABLE signalmeow_contacts RENAME COLUMN profile_fetch_ts TO profile_fetched_at;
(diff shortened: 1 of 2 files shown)


tulir:
I thought EMS used Postgres 🤔

AndrewFerr:
> I thought EMS used Postgres 🤔

It does, but my local dev environment does not :slightly_smiling_face:

tulir:
You could just manually fix it :P The v7 -> v8 migration is exclusively for users of your PR, the main branch jumped from v6 to v8 and the alter tables there are already scoped to postgres only (because they're less important).

Anyway, if you really want sqlite support for that migration: there are no foreign keys referencing `signalmeow_contacts`, so disabling foreign keys and transactions isn't necessary and therefore a normal dialect-split upgrade works perfectly fine. The way to do that is to have two files with the same name and header comment, but one file name ending in `.sqlite.sql` and the other in `.postgres.sql`. https://github.com/mautrix/discord/blob/main/database/upgrades/13-merge-emoji-and-file.sqlite.sql + https://github.com/mautrix/discord/blob/main/database/upgrades/13-merge-emoji-and-file.postgres.sql is an example of that

AndrewFerr:
Thanks for the tips!  I wasn't aware of the file-extension-based dialect splitting.  Part of my reasoning in writing this PR is to learn the "proper" way to do mautrix schema upgrades, so this is a good outcome.

Issue #476, opened 2024-03-12
Signal profile avatar changes may not appear in Matrix


AndrewFerr:
If a Signal user changes their avatar, the new avatar might not appear in Matrix.

Part of why this happens is because `pkg/signalmeow/profile.go` returns [any cached profile from the last hour](https://github.com/mautrix/signal/blob/2f7d81f97fd384b0ae00266e50756840ce5f5620/pkg/signalmeow/profile.go#L123) instead of always fetching the latest profile.  This suggests that a Signal avatar change will appear in Matrix as long as the avatar is changed at least an hour after the last time it was changed.

A quick fix is to not use cached profiles, but that would obviously increase the number of profile fetches.

maltee1:
I don't think there's a way to not cache profiles, since profiles are essentially being cached as avatar and displayname on matrix. There isn't really any way for the bridge to tell whether the matrix user is currently looking at a signal ghost's profile, to fetch it at that moment.

PR #478, opened 2024-03-12
Reduce profile expiry time

AndrewFerr:
For #476

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -214,7 +214,7 @@ func fnPM(ce *WrappedCommandEvent) {
 	user := ce.User
 	var targetUUID uuid.UUID
 
-	if contact, err := user.Client.ContactByE164(ce.Ctx, fmt.Sprintf("+%d", number)); err != nil {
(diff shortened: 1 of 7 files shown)


tulir:
Probably better to just wait an hour for avatars to update 🤔

AndrewFerr:
A compromise could be to add a bridge config option to set the profile expiry time.

The situation I wanted to avoid is someone changing their Signal avatar a few times & a Matrix user being unaware of it until later.

tulir:
Yeah making it configurable is probably fine

AndrewFerr:
Actually, a better way is to fetch the profile when receiving a `SyncMessage` with a `FetchLatest` indicating a profile change.  Then the existing fetch can remain as-is, but used as a safety check in case an on-demand was missed.

AndrewFerr:
Pushed the config change for posterity, but abandoning this in favour of #482

AndrewFerr:
It's been pointed out that the `SyncMessage` for profile changes is only for logged-in users, so it's not a full solution as I had thought.

Still, I'll leave this PR closed because reducing the profile expiry time incurs the risk of doing a lot of extra fetches, as it would trigger a fetch on _any_ new Signal message (plus some other Signal events) that happen after the expiry time, even if the profile hasn't changed.

PR #482, opened 2024-03-13
Add handler for profile changes

AndrewFerr:
This causes Signal profile changes to be fetched as they happen, but only for logged-in users.

Helps with #476


tulir:
Signal actually sends events for profile changes? I couldn't see anything when I last tried it 🤔 (or is this only for the logged-in user?)

PR #485, opened 2024-03-17
Pass arguments to docker-run.sh on to mautrix-signal

### LINKED ISSUE #474: Pass arguments to `docker-run.sh` on to mautrix-signal
The `mautrix-signal` binary provides a few command line flags to modify its behavior, most recently `--ignore-unsupported-server` as a workaround to have it function with Dendrite (at least its current `main` branch). However, setting these options when using the Docker image is not very straightforward. One would have to override the image's entire `CMD` to explicitly invoke `mautrix-signal`, but that also means losing the registration generation and such that `docker-run.sh` usually takes care of.

A solution for this would be have the `docker-run.sh` script pass any additional arguments on to the mautrix-signal invocation within it. Additionally, it may make sense to invoke the script via `ENTRYPOINT` instead of `CMD` so that the latter is used entirely for user-provided arguments, but even being able to override `CMD` as e.g. `["/docker-run.sh", "--ignore-unsupported-server"]` would suffice.

Alternatively, mautrix could also provide environment variables for setting these options.

BVollmerhaus:
This PR extends `docker-run.sh` so that it passes additional arguments on to mautrix-signal (once the requisite files exist and the bridge is started). In particular, this should make it easier to use CLI flags such as `--ignore-unsupported-server` when using the Docker image (by overriding `CMD`).

Further, it fixes some inconsistent indentation and ShellCheck warnings concerning missing variable quoting.

Closes #474

### DIFF
--- a/docker-run.sh
+++ b/docker-run.sh
@@ -6,7 +6,7 @@ fi
 
 # Define functions.
 function fixperms {
-	chown -R $UID:$GID /data
+	chown -R "$UID:$GID" /data
 
 	# /opt/mautrix-signal is read-only, so disable file logging if it's pointing there.
 	if [[ "$(yq e '.logging.writers[1].filename' /data/config.yaml)" == "./logs/mautrix-signal.log" ]]; then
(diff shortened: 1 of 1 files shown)


BVollmerhaus:
@tulir Hey, could you have a look? :slightly_smiling_face: I think this would be quite useful for Dendrite users in particular.

PR #488, opened 2024-03-24
Handle missed GroupChanges using group history

maltee1:
I was hoping fetching group history would help with correct metadata during portal creation but (obviously, probably) it only allows you to fetch revisions from when a user was added.

Might still come in handy for more than just handling missed group changes, which shouldn't ever happen anyway.

### DIFF
--- a/pkg/signalmeow/groups.go
+++ b/pkg/signalmeow/groups.go
@@ -178,6 +178,7 @@ type BannedMember struct {
 type GroupChange struct {
 	groupMasterKey types.SerializedGroupMasterKey
 
+	SourceACI                          uuid.UUID
 	Revision                           uint32
 	AddMembers                         []*AddMember
 	DeleteMembers                      []*uuid.UUID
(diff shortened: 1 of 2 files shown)


pkg/signalmeow/groups.go
```diff
@@ -176,6 +176,7 @@ type BannedMember struct {
 type GroupChange struct {
 	groupMasterKey types.SerializedGroupMasterKey
 
+	SourceServiceId                    uuid.UUID
```
tulir, line 179: This looks wrong (name doesn't match type and `Id` should be `ID`)

maltee1: Good point, now should be better.

portal.go
```diff
@@ -921,6 +930,30 @@ func (portal *Portal) handleSignalDataMessage(source *User, sender *Puppet, msg
 	}
 }
 
+func (portal *Portal) catchUpHistory(source *User, fromRevision uint32, toRevision uint32, ts uint64) error {
+	log := portal.log.With().
+		Str("action", "catchUpHistory").
+		Stringer("source", source.MXID).
+		Uint32("from revision", fromRevision).
+		Uint32("to revision", toRevision).
```
tulir, line 938:
```suggestion
		Uint32("from_revision", fromRevision).
		Uint32("to_revision", toRevision).
```

log keys shouldn't have spaces

tulir (approved):

PR #489, opened 2024-03-24
Sync power levels and join rule when creating matrix room

maltee1:


portal.go
```diff
@@ -1772,11 +1772,10 @@ func (portal *Portal) CreateMatrixRoom(ctx context.Context, user *User, groupRev
 		return err
 	}
 	portal.log.Info().Msg("Created matrix room for portal")
-
+	members := portal.SyncParticipants(ctx, user, groupInfo)
+	portal.updatePowerLevelsAndJoinRule(ctx, groupInfo, members)
```
tulir, line 1776: The power levels and join rule should probably be set in initial state rather than after creating 🤔

maltee1: Tricky. Power levels are synced using the members map that `SyncParticipants` returns. That one skips pending members and the source user. The latter can probably be added later, the former only with significantly more effort. Of course, we could also not sync power levels for pending members, but since `PromotePendingMembers` doesn't come with a `Role`, we would have to fetch full group info every time a pending member is promoted.

I suppose splitting power levels and join rule into two functions, with the join rule being added in the initial state would work without much effort.

Does it really make that much of a difference whether power levels are part of the initial state or synced a second later?

maltee1:
rendered obsolete by #501

PR #496, opened 2024-04-05
Minor incoming messages refactor and send m.notice on decryption errors

smweber:
We were falling through without propagating errors anywhere. I caught most decryption errors and send them along in the new Err field in DecryptionResult, to be then passed along to the bridge as a new type of chat event.

I'm going to do another pass to catch any last decryption errors, but this should be a big improvement over what we currently do.


tulir (approved):
The diff is impossible to review, but the new code looks fairly sensible

PR #498, opened 2024-04-11
try fix zero length UserId error

maltee1:
So far untested, need tester with corresponding error (doesn't seem to be causing new breakage)


maltee1:
[mautrix-signal.zip](https://github.com/mautrix/signal/files/14950876/mautrix-signal.zip)
for those who don't want to compile

PR #506, opened 2024-04-23
Properly convert signal long text messages

### LINKED ISSUE #479: Long messages are sent as `m.file` with part of the message as `body`
I suppose signal sends long messages as attachments with a preview text in the caption. The attached file is `"mimetype": "text/x-signal-plain"`. It contains the full message contents.

The bridge should probably look for that mimetype and send a regular message with the downloaded contents instead.

Mimicking that behavior for Matrix -> signal messages does not seem to be necessary. I sent one of those attachment messages in full from matrix and it was bridged just fine.

maltee1:
Fixes #479


tulir (approved):
Looks like it might work, although I think there were also some edge cases with text formatting

maltee1:
Formatting works with this

PR #516, opened 2024-05-30
Fix example location URL template config for OSM

AndrewFerr:


sumnerevans (approved):

sumnerevans (commented):
Actually, I don't think this really has any effect. I tried out a bunch of URLS, and openstreetmap is very permissive when it comes to the format of the latitude and longitude. the `s` suffix was not necessary on either `mlat` or `mlon`.

PR #527, opened 2024-08-08
handle Membership

maltee1:
matrix -> signal membership. Missing are knocks and joins, as they need access to groupmasterkeys through invite links, which is a little more complicated. Those will follow later.


pkg/signalid/ids.go
```diff
@@ -39,10 +39,25 @@ func ParseUserID(userID networkid.UserID) (uuid.UUID, error) {
 	}
 }
 
+func ParseUserLoginID(userLoginID networkid.UserLoginID) (uuid.UUID, error) {
+	serviceID, err := ParseUserLoginIDAsServiceID(userLoginID)
+	if err != nil {
+		return uuid.Nil, err
```
tulir, line 45: UserLoginIDs are always ACIs

pkg/connector/handlematrix.go
```diff
@@ -344,3 +346,107 @@ func (s *SignalClient) HandleMatrixRoomTopic(ctx context.Context, msg *bridgev2.
...
+		gc.PromoteRequestingMembers = []*signalmeow.RoleMember{{
+			ACI:  targetSignalID,
+			Role: role,
+		}}
+	case bridgev2.RetractKnock, bridgev2.RejectKnock, bridgev2.BanKnocked:
+		gc.DeleteRequestingMembers = []*uuid.UUID{&targetSignalID}
+	case bridgev2.Unban:
+		unbanUser := libsignalgo.NewACIServiceID(targetSignalID)
+		gc.DeleteBannedMembers = []*libsignalgo.ServiceID{&unbanUser}
+	}
```
tulir, line 434: This should probably have a default case that returns without doing anything

pkg/connector/handlematrix.go
```diff
@@ -344,3 +346,107 @@ func (s *SignalClient) HandleMatrixRoomTopic(ctx context.Context, msg *bridgev2.
...
+			}
+			targetIntent = ghost.Intent
+		}
+	}
+	gc := &signalmeow.GroupChange{}
+	role := signalmeow.GroupMember_DEFAULT
+	toMembership := msg.Type.GetTo()
+	if toMembership == event.MembershipBan {
+		gc.AddBannedMembers = []*signalmeow.BannedMember{{
+			ServiceID: libsignalgo.NewACIServiceID(targetSignalID),
```
tulir, line 379: Why is this case separately from the switch 🤔

maltee1: I separated Bans into different membership transitions, depending on what the previous membership is. Those remove the user from the group, remove an invite etc. I could add the BannedMember each time, but I figured it would be easier to only do that once.

tulir: You can have a `case` with multiple values (`case bridgev2.Ban, bridgev2.BanJoined, bridgev2.BanInvited: ...`)

maltee1: I know that, but I would then have to separate out the different from memberships at that point - I don't see the advantage over what I'm currently doing.

tulir: The advantage is having a single switch statement that covers every membership change, which is cleaner and allows having a `default` block for unsupported changes

pkg/signalid/ids.go
```diff
@@ -39,10 +39,22 @@ func ParseUserID(userID networkid.UserID) (uuid.UUID, error) {
 	}
 }
 
+func ParseUserLoginID(userLoginID networkid.UserLoginID) (uuid.UUID, error) {
+	userID, err := ParseUserLoginID(userLoginID)
```
tulir, line 43:
```suggestion
	userID, err := uuid.Parse(string(userLoginID))
```

pkg/signalid/ids.go
```diff
@@ -39,10 +39,22 @@ func ParseUserID(userID networkid.UserID) (uuid.UUID, error) {
...
+}
+
 func ParseUserIDAsServiceID(userID networkid.UserID) (libsignalgo.ServiceID, error) {
 	return libsignalgo.ServiceIDFromString(string(userID))
 }
 
+func ParseUserLoginIDAsServiceID(userLoginID networkid.UserLoginID) (libsignalgo.ServiceID, error) {
+	return libsignalgo.ServiceIDFromString(string(userLoginID))
+}
+
```
tulir, line 57:
```suggestion
```

pkg/connector/handlematrix.go
```diff
@@ -25,11 +25,13 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/rs/zerolog"
+	"github.com/rs/zerolog/log"
```
tulir, line 28: This also has a global log somewhere

pkg/signalid/ids.go
tulir, line 57: I think your force push reverted these fixes

tulir (approved):
Otherwise looks good

PR #501, opened 2024-04-19
Full sync on portal creation, invite handling

maltee1:
- refactor SyncParticipants and updatePowerLevelsAndJoinRule to
- set power levels and join rule in initial state
- sync pending/requesting/banned members on room creation
- bridge room creation with actual group creator, if known
- support identifying logged-in users by PNI
- support for group invites, accepting or rejecting invites
- changed SourceACI to SourceServiceID; PNI type SourceServiceIDs seem to exist - not sure why.

Some oddities: if the group creation is double-puppeted, the bridge bot will say "Hi, I'm a Signal bridge bot". Not sure how to turn that off. Group creation by ghost is not a problem
Sometimes the bridge bot will double down on group creation invites

PR #528, opened 2024-08-08
Look up ACIs for PNIs

maltee1:
I wasn't able to test this due to lack of new contacts, but it's mostly copied over from #501  so it should work all the same

### DIFF
--- a/pkg/connector/groupinfo.go
+++ b/pkg/connector/groupinfo.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/rs/zerolog"
 	"go.mau.fi/util/ptr"
 	"maunium.net/go/mautrix/bridgev2"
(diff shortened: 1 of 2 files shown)


pkg/connector/groupinfo.go
```diff
@@ -120,11 +121,18 @@ func (s *SignalClient) getGroupInfo(ctx context.Context, groupID types.GroupIden
 		}
 	}
 	for _, member := range groupInfo.PendingMembers {
-		if member.ServiceID.Type != libsignalgo.ServiceIDTypeACI {
-			continue
+		var aci uuid.UUID
+		if member.ServiceID.Type == libsignalgo.ServiceIDTypePNI {
+			device, err := s.Client.Store.DeviceStore.DeviceByPNI(ctx, member.ServiceID.UUID)
+			if err != nil {
```
tulir, line 127: I think all of these cases need to check for `device == nil` too

maltee1:
On a second thought, I should probably write a function instead of boilerplate.

pkg/signalmeow/store/container.go
```diff
@@ -39,6 +40,7 @@ FROM signalmeow_device
 `
 
 const getDeviceQuery = getAllDevicesQuery + " WHERE aci_uuid=$1"
+const deviceByPNIQuery = getAllDevicesQuery + "Where pni_uuid=$1"
```
tulir, line 43:
```suggestion
const deviceByPNIQuery = getAllDevicesQuery + "WHERE pni_uuid=$1"
```

tulir (approved):

PR #529, opened 2024-08-12
handle PromotePending(PniAci)Members

maltee1:
I noticed this was missing but I can't test now due to lack of new contacts.

Is there anything that can go wrong?


tulir (approved):
Oh, promote means accept invite, signal is weird

maltee1:
Apparently PromoteRequestingMembers was also missing. Does that work, going from `knock` to `join`?

tulir (approved):

PR #537, opened 2024-08-20
Skip unnecessary membership changes when banning

maltee1:


pkg/connector/groupinfo.go
```diff
@@ -261,9 +278,14 @@ func (s *SignalClient) groupChangeToChatInfoChange(ctx context.Context, rev uint
 		if aci == nil {
 			continue
 		}
+		membership := event.MembershipLeave
+		if _, ok := bannedMembers[*memberServiceID]; ok {
+			membership = event.MembershipBan
+			delete(bannedMembers, *memberServiceID)
```
tulir, line 284: Seems like it'd be easier to just ignore the other changes when there's a ban, it probably doesn't need to check the prev membership

tulir (approved):

PR #600, opened 2025-05-16
directmedia: actually return properly decoded profile avatar info

adamvy:


adamvy:
this appears to work on initial sync when connecting the bridge.

However avatar/name changes don't seem to be picked up, do we actually refresh to ghosts anywhere?

adamvy:
did eventually come through it seems, but wasn't on next message

PR #608, opened 2025-09-25
resync: resync portals upon viewieng if they haven't synced in the last 24h

adamvy:


pkg/signalid/dbmeta.go
```diff
@@ -23,6 +23,9 @@ import (
 type PortalMetadata struct {
 	Revision               uint32 `json:"revision,omitempty"`
 	ExpirationTimerVersion uint32 `json:"expiration_timer_version,omitempty"`
+	// Lazy resync tracking
+	LastSync          jsontime.Unix `json:"last_sync,omitempty"`
+	BridgeCapsVersion int           `json:"bridge_caps_version,omitempty"`
```
tulir, line 28: This looks unused (and portals have a CapState field already anyway)

```suggestion
```

PR #623, opened 2025-11-26
signalmeow/storageservice: fix nickname clearing not being bridged

### LINKED ISSUE #622: Clearing nickname not propagated
When I update a contact's nickname in the Signal app, their Matrix account name changes immediately. However, if I delete their nickname in the Signal app, the Matrix account continues to remain as their former nickname.

I'm running mautrix-signal v0.2511.0

Enzime:
Fixes #622


Enzime:
@tulir I'm using the Dockerfile to build the Signal bridge and it seems like the Dockerfile is broken on `main`, should I include the fixes in this PR or a separate PR?

PR #649, opened 2026-05-12
capabilities: drop webp to only partial support

SpiritCroc:
Signal doesn't animate webp, so clients may chose to prefer APNGs instead.

### Checklist

* [x] I have read and followed the contributing guidelines at <https://docs.mau.fi/bridges/general/contributing.html>

### DIFF
--- a/pkg/connector/capabilities.go
+++ b/pkg/connector/capabilities.go
@@ -75,8 +75,7 @@ var signalCaps = &event.RoomFeatures{
 				"image/gif":  event.CapLevelFullySupported,
 				"image/png":  event.CapLevelFullySupported,
 				"image/jpeg": event.CapLevelFullySupported,
-				// Signal clients will only render static webp, so apng is preferred
-				"image/webp": event.CapLevelPartialSupport,
+				"image/webp": event.CapLevelFullySupported,
 				"image/bmp":  event.CapLevelFullySupported,
 			},
 			MaxWidth:         4096,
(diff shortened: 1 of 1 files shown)


pkg/connector/capabilities.go
```diff
@@ -111,7 +111,7 @@ var signalCaps = &event.RoomFeatures{
 		},
 		event.CapMsgSticker: {
 			MimeTypes: map[string]event.CapabilitySupportLevel{
-				"image/webp": event.CapLevelFullySupported,
+				"image/webp": event.CapLevelPartialSupport,
```
tulir, line 115: We might want to define more precise signaling later, so these should be documented

```suggestion
				// Signal clients will only render static webp, so apng is preferred
				"image/webp": event.CapLevelPartialSupport,
```

pkg/connector/capabilities.go
```diff
@@ -75,7 +75,8 @@ var signalCaps = &event.RoomFeatures{
 				"image/gif":  event.CapLevelFullySupported,
 				"image/png":  event.CapLevelFullySupported,
 				"image/jpeg": event.CapLevelFullySupported,
-				"image/webp": event.CapLevelFullySupported,
+				// Signal clients will only render static webp, so apng is preferred
+				"image/webp": event.CapLevelPartialSupport,
```
tulir: Not sure if it's correct for `m.image`, those probably shouldn't have animation anyway and apng isn't supported (gif is a separate fake msgtype below)

tulir (approved):

PR #657, opened 2026-07-21
login: implement proper error handling

Fizzadar:
Updated based on the Signal server code for the `PUT /v1/devices/link` endpoint:

https://github.com/signalapp/Signal-Server/blob/6a8bf7f78e3516421382f7773762bf2f7f0a78a9/service/src/main/java/org/w
  hispersystems/textsecuregcm/controllers/DeviceController.java

Also bumped up the QR limit to 15 minutes because 4.5 mins to fetch device -> signal -> settings -> link device is tight unless you've done it before.

### DIFF
--- a/pkg/connector/login.go
+++ b/pkg/connector/login.go
@@ -18,9 +18,12 @@ package connector
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net/http"
 	"time"
 
+	"github.com/coder/websocket"
 	"github.com/google/uuid"
 	"maunium.net/go/mautrix/bridgev2"
 	"maunium.net/go/mautrix/bridgev2/database"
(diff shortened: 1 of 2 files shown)


pkg/connector/login.go
```diff
@@ -70,6 +73,72 @@ const (
...
+		case http.StatusConflict:
+			return ErrDeviceLinkMissingCapability
+		case http.StatusLengthRequired:
+			return ErrDeviceLimitReached
+		case http.StatusForbidden:
+			return ErrDeviceLinkCodeInvalid
+		case http.StatusTooManyRequests:
+			return ErrDeviceLinkRateLimited
+		default:
+			return ErrDeviceLinkRejected.AppendMessage(" (HTTP %d)", linkErr.StatusCode)
```
tulir, line 136: This seems to be throwing away the message from the server, does that never contain useful info?

tulir (approved):
Otherwise looks fine

PR #661, opened 2026-08-03
handle*: insert stub message rows for old edit events

Fizzadar:
This prevents a re-delivery of an old edit breaking the bridge due to duplicate network ID error as the original message row will be updated by the edit.

### DIFF
--- a/pkg/connector/handlesignal.go
+++ b/pkg/connector/handlesignal.go
@@ -392,8 +392,11 @@ func (evt *Bv2ChatEvent) ConvertEdit(ctx context.Context, portal *bridgev2.Porta
 	converted := evt.s.Main.MsgConv.ToMatrix(ctx, evt.s.Client, portal, evt.Info.Sender, intent, editMsg.GetDataMessage(), nil)
 	// TODO can anything other than the text be edited?
 	editPart := converted.Parts[len(converted.Parts)-1].ToEditPart(existing[len(existing)-1])
-	editPart.Part.EditCount++
 	prevID := editPart.Part.ID
+	// Clone the database message struct to avoid mutating the ID.
+	// The ID from the original struct is used for AddedParts (we specifically want the old ID for that)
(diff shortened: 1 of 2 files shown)


pkg/connector/handlesignal.go
```diff
@@ -358,17 +359,54 @@ func (evt *Bv2ChatEvent) ConvertMessage(ctx context.Context, portal *bridgev2.Po
...
 	// TODO tell converter about existing parts to avoid reupload?
 	converted := evt.s.Main.MsgConv.ToMatrix(ctx, evt.s.Client, portal, evt.Info.Sender, intent, editMsg.GetDataMessage(), nil)
 	// TODO can anything other than the text be edited?
 	editPart := converted.Parts[len(converted.Parts)-1].ToEditPart(existing[len(existing)-1])
 	editPart.Part.EditCount++
+	prevID := editPart.Part.ID
 	editPart.Part.ID = signalid.MakeMessageID(evt.Info.Sender, editMsg.GetDataMessage().GetTimestamp())
+	if prevID != editPart.Part.ID {
+		saveEditStub(ctx, evt.s.Main.Bridge, prevID, editPart.Part)
+	}
```
tulir, line 406: Hmm, both these calls will probably fail because the original message hasn't been edited yet.

For Matrix handling, it should be fine to just call `bridge.DB.Message.Update` before saving the new stub. bridgev2 will call Update again, but that's basically a no-op then.

For Signal handling, it probably has to either go in PostHandle or maybe as an item in AddedParts with DontBridge set. I think AddedParts automatically uses the old message ID too

Fizzadar: Updated!

tulir: I lied, AddedParts uses `existing[0].ID`, which is mutated here

Added a clone so it doesn't get mutated

tulir (approved):

PR #664, opened 2026-08-14
signalmeow/storageservice: store own profile key from account record

Fizzadar:
The bridge's own profile key is only written at link time, so it goes stale once the primary device rotates it, which breaks expiring profile key credential fetches and therefore group creation.

Adopt the profile key from the storage service account record, like contact records already do.


tulir (approved):
Profile keys are rotated? 🙀

Fizzadar:
> Profile keys are rotated? 🙀

Apparently so! First time we've ever seen it, maybe device reset or something something?

PR #665, opened 2026-08-19
signalmeow/web: use fresh request time when retrying

Fizzadar:
When a websocket stops responding, the pending request is retried after the connection is torn down, but it kept the original request time and was therefore rejected as too old by the new connection's write loop.

### DIFF
--- a/pkg/signalmeow/web/signalwebsocket.go
+++ b/pkg/signalmeow/web/signalwebsocket.go
@@ -616,13 +616,12 @@ func (s *SignalWebsocket) SendRequest(
 		Path:    &path,
 		Body:    body,
 		Headers: headerArray,
-	}, time.Now(), 0)
+	}, 0)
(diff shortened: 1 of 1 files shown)


pkg/signalmeow/web/signalwebsocket.go
```diff
@@ -657,7 +657,7 @@ func (s *SignalWebsocket) sendRequestInternal(
 			}
 		}
 		zerolog.Ctx(ctx).Warn().Int("retry_count", retryCount).Msg("Received nil response, retrying recursively")
-		return s.sendRequestInternal(ctx, request, startTime, retryCount+1)
+		return s.sendRequestInternal(ctx, request, time.Now(), retryCount+1)
```
tulir: Should the parameter just be removed entirely if it's always `time.Now()`?

